PR #11, opened 2024-03-27
readme updated

lmangall:
from https://linear.app/web-serv/issue/WEB-16/test-issue-for-leo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,71 @@
-# webserver
 
-Our own HTTP server!
+
+## Last resources I have been checking out (Leo's selection):
+
+- I used a cached web archive otherwise you have to pay for Medium
+  [HTTP Server: What do you need to know to build a simple HTTP server from scratch](https://web.archive.org/web/20190124234346/https://medium.com/from-the-scratch/http-server-what-do-you-need-to-know-to-build-a-simple-http-server-from-scratch-d1ef8945e4fa)
+
+- [GitHub - Kaydooo/Webserv_42](https://github.com/Kaydooo/Webserv_42?tab=readme-ov-file)
+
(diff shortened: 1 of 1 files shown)


dantol29 (approved):
That is a real README!🚀

552020:
Pity I've not been selected as a reviewer!  🙁

lmangall:
@552020  As far as I know you are. 2 review are required and only yours now is missing. 
Or is there a way to explicitly write somebody as a reviewer ?

lmangall:
Alright, saw your message in #12 and added you as a reviewer

552020 (changes requested):
How strict should I be with my review? Or being this a test issue, should I pass everything?

My critic would be that this being the README of the project should give first the basic information about the project itself, style should be impersonal and brief (Resources instead of "Last resourced I have been checking out (Leo's selection)' and not out of context "Allowed functions". 

But if you mean it's just about completing the linear workflow and we don't care about having a well curated README I can pass approve the MR without further modifications. ;-)

lmangall:
The Linear was "Add something into the README" si I did just that. 
This document is the very first thing we see when we open the page ("we" because nobody else is going to open it, especially before it's done). So while in development I personnaly treat the Readme as notes on the repo and as an help for myself rather than for people visiting the page. Which is indeed a point I didn't made clear.
I think it is useful to collect links or important information we might want to check or verify while we develop this application. That is why I had "Last resourced I have been checking out (Leo's selection)'" : you both can add your own list so if the others want to check it they can, or you can add the crucial information you think should be present in the first page of the repo when we open it. And perhaps when the project is done we can turn this Readme into a form that fulfils its conventional purpose.
I even think the Readme should be the exception to the "no push to main" rule. When you open the repo you have the Readme and on its right a pencil icon so you can edit it right on, really easily...

552020 (approved):
LGTM

PR #13, opened 2024-03-27
added stefano s documentation to the main repo

dantol29:
Resources.md file has to be moved to the readme.md, since we store links there.

I created 'stefano_sockets' folder to separate different implementations.

Added docs folder with lots of files from Stefano's repo.


lmangall (changes requested):
I think we don't need the code (.py, .cpp and MAKEFILE) the README.md is enough

dantol29:
I deleted all the code, now should be fine

lmangall (approved):
Thanks, my original post was before I saw that the code was related to another "issue"/ task 
There is a lot of useful information in those docs !

docs/socket.md
552020, line 1: The article about socket is just a stub and it should not be "copied" like that into main I would say. Did you @lmangall read and approve it?

I can improve it if you think that it is of any utility to have a document like this in our repo, otherwise I can also approve the PR the way it is.

552020 (changes requested):
I asked some question about the utility of having a 'stub' like socket.md into main, with just copy/pasted content. 

I also make some changed to some articles, which were not precise, like in the possible usage of sockaddr_in there were functions listed, which we can't use and I added an extra fie about the family of the sockaddr_in structures with some notes about the proabability we are going to use them in our project. 

For @dantol29: what I did (pushing directly on your branch) is not a good practice, since this is your branch, and you don't want other people changing stuff in your branch. I should have chosen one of the two possibilites:
1. ask you to make the changes
2. open a new branch from your branch, make the changes I thought were necessary and the open a PR for my branch to your branch. 

But since this was documentation that I wrote and you took it basically in the form it was, I allowed myself to change it directly :-)

docs/socket.md
lmangall: I didn't read carefully, sorry. Just went over it. I still am not sure what is wrong with the article, you mean it's copied and unedited ?
It's nice to have something, but at the moment a couple lines are enough, no need to do a full documentation...

552020 (approved):
LGTM

552020:
Since there were changes you need to approve it again for it to be merged @lmangall

lmangall (approved):
Enough lecture for the weekend !

PR #17, opened 2024-03-28
Display something on the browser

552020:
- Lil modification on the mini_websserv in my repo, which displays some html. 
- Usage is explained in the README
- Some docs have been added
- Also some stuff in the .vscode directory: this are the settings I have in my cpp projects, if you have other settings we can agree on it, or not commit the .vscode dir at all (which was already committed)
- I add also a .clang_format file, for the formatting of the code: here is important the we find a common formatting style so that we don't get modifications in git cause of formatting stuff, which makes code reviews much harder.


dantol29:
Great documentation! Man.cpp totally makes sense!

.vscode/c_cpp_propertied.json
lmangall, line 1: Can you explain how this json is being used ? especially :
      "cppStandard": "c++98",

lmangall (approved):
Thanks for the documentation. I heard of Postman and it seems to be a tool one has to master, have you ever used it ?

.vscode/c_cpp_propertied.json
552020: I guess it tells to the language server of VSCode that the C++ version being used is C++98, so it should underline in red if you use features which are not part of C++98 (like auto). But it never happend to me that I got this warning.

552020:
> Great documentation! Man.cpp totally makes sense!

Thanks! @dantol29! What about approving it? :-)

552020:
> Thanks for the documentation. I heard of Postman and it seems to be a tool one has to master, have you ever used it ?

@lmangall I did in many Udemy courses when you need to test your application. It's pretty easy to use. You can download it and make the same request you did through the browser to our server with postman: you just put the same address and you will see the response. You can get the same information also in the dev tool of the browser (like the time of the response and the content of the header and body of the http request). But in postman you can save different requests and you can perform them programmatically.

dantol29 (approved):
Great!

PR #25, opened 2024-03-30
Lmangall/web 24 cgi to the browser

lmangall:
Some of the work/steps I have documented in [the slack message](https://webservinc.slack.com/archives/C06R2QT1X4L/p1711718692897499) and looking at the commit history can be helpful.

In the end, Stefano's orgiginal html code returned from main became a home.html file that is "read" when one is on the "home" of our Webserv with the help of `readFile()`

In the last commits I split up the code, now the while loop of the main uses  `handleConnection(int socket) ` to call one of these:
```
void handleHomePage(int socket) 
void handleHelloPage(int socket) 
void handleNotFound(int socket) 
```
I also created a .hpp in order to be able to add doxygen comments (which should show as in the picture below) and not have them "in the way"

Screenshot 2024-03-30 at 10 24 21


lmangall:
My next step is WEB-29 : https://linear.app/web-serv/issue/WEB-29/cgi-pass-a-query-string-to-the-cgi

552020:
Thanks! Remember to add the reviewers, otherwise we don't get notified! :-)

552020 (commented):
0. I managed to found out how to use the cgi script and http://localhost:8080/hello works! Greate job!

1. Are you using the not allowed `popen` (never heard of it) function instead of forking on purpose? 

2. Are handleHomePage and handleNotFound also part of the CGI functionalities?

3. http://localhost:8080/home and http://localhost:8080/ are not working, and I guess the code should have a ./home.html file to be read somewhere (if I understand the code) file but there is not such file.

4. Is there any particular reason for you to deleted my code for rendering http://localhost:8080/?

552020:
Point 3 is for sure cause of .gitignore

lmangall:
Sorry for the reviewers, I keep forgetting.
1.  popen() use is accidental, I corrected it, maybe there is a better way to do it

2.  handleHomePage and handleNotFound are not part of the CGI functionalities. First I modified http://localhost:8080/ to show the cgi, then I moved the cgi to its own page, so at that point http://localhost:8080/ looked broken and I felt bad about taking off your code so I put it back with a small improvement (because if I understood well in the end this kind of pages should be .html in our webserver) and then handleNotFound was  then next the associated little improvement

3.  Both http://localhost:8080/home and http://localhost:8080/ (should) work, but I forgot to precise (maybe I did in the last PR) that the file `/cgi-bin/hello.cpp` has to be compiled (with `-o hello.cgi`). Maybe it could be done has an option in the makefile. So yes, it doesn't work "right off". I think it is better practice to have it compiled in the makefile as an option than to modify the .gitignore and push the executable but we can discuss it.

4. As I wrote I didn't deleted it,

> I wouldn't dare !

 only made it an html page, which is served when the user request http://localhost:8080/home or http://localhost:8080/  but probably you missed because of the compilation

lmangall:
I have added the compilation of `CGI_SRC = cgi-bin/hello.cpp` in the makefile, at the moment not as a flag. We can modify it later but at least the program is compilable and usable right off !

552020:
Short answer during eastern lunch: try a fresh clone of the project. The .html are gitignored, cause everything is. You need to add something like !*.html 

Real cgi was working and I precompiled it. Instructions were in the bottom of the file. To add it into Makefile is a good idea

lmangall:
Thanks, I thought that the .gitignore caused a problem only to the executable.

I modified the .gitignore, and also the Makefile, I tried on a new clone and now it should be perfect now !

Short answer but with just the right info.
Enjoy easter !

552020 (commented):
1. - .gitignore: everything fine: .cgi-bin/ seems to be the best-practice directory to put the binaries of the cgi script. Do you think that is also ok to put the script itself in the a '-bin' directory? Can you check if it's a best practice also to put the source code for the cgi script in a 'bin' directory?

2. - Makefile: I was going to thank you for having spotted the wrong compiler `c++` instead of `g++`, but I checked the subject again and I found this!

image

3. cgi-bin/hello.cpp

Why do we use the forbidden getenv() in the script?

4. README_CGI

4. A Markdown files start normally with a `#title`
4. B Can we add a line about the difference between 'scripts' and 'program'. Cause I keep on forgetting it.

4. C CGI scripts are conventionally stored in `/cgi-bin/`, and have a `.cgi` extension. 
`cgi-bin` or `.cgi-bin`. Hidden or not?
4. D The good rfc
>> ---(IS THAT THE GOOD RFC ?)-

What do you mean?

Otherwise good!


- Webserv_2.pdf

I'm kind of against pdf or any not "readable" content in the repo. But this is secondary. The name of the pdf could be more meaningful just to know what is about. The best solution would be to have it as Markdown and if a diagram is necessary then I would add it as .jpg just to make sure that it's not textual content that could have been displayed in another way. (but this is more an 'aesthetical' issue, with no priority. 

- home.html. I checked now how nginx solves the issues of 'hard-coded' pages. It has a couple of .html pages in an /html directory. Maybe we could follow that pattern. If you check in my repo, I copied the source code of ngnix. There is also a Dockerfile to run it in an ubuntu environment. We are supposed to take ngnix as reference. The .html file instead of the HTML in a functioni (like I add it) is in any case a step forward. I guess I would just copy the HTML from ngnix, as a 'citation' and change with our data. But feel free to change is the way you want.

552020 (commented):
To the handleHelloPage, this is the function that handles the cgi functionality, maybe it makes sense to give to it a name that goes in that direction. Not the function is taking the file descriptor of a socket and it's writing into it directly with write. Is this the way it shoud work or it shoud take the http request (and eventually the socket id and give back the http response and the socket id). I didn't delve into the question: so this is a real answer. We are not allowed to write like that into sockets but every read/write should be behind a poll function. So probably is better to take int socket and unprocessed content and give back int socket and processed content. 

The handleConnection is a sort of router (before the cgi). We can write an extra issue for it and start to think about how it shoudl work cause I have no idea how to design it in the context of the server core functionalities.

lmangall:
1.  The source code for the script should be "kept away" ideally in a place not accesible from the web for security reasons (perhaps this is a perfect use for your secret directory)

2. g++ instead of c++ is probably a collateral change when I was trying to have the Nakefile to compile the script source code

3. I think the CGI executed program is exempt of the limitation from the subject. It is not per se a part of our webserver and the work we have to deliver, only an addition to demostrate how it works. 
In this case I thought the code could take a string query, and I modified the main so it passes an hardcoded string query through execve. I think our webserv should send even more environment variables to the CGI executed script. Actually I now think it would be better to have a php scrit instead of a compiled C++ binary, perhaps I will do that next

4. README_CGI 
- now has a markdown formatted title
- now explain the difference between programs and scripts (In a nutshell: scripts are interpreted at runtime, programs are compiled)
- the good RFC: everything was ok, I removed this comment to myself



regarding the pdf: I took it out, it is already outdated, it was more intended as an internal (between us) note


Should the cgi-bin directory be hidden ?
The Friend:
By default, the cgi-bin directory is not "hidden" in the sense that web users can request and execute the CGI programs contained within it via their web browser. This is the directory's primary purpose: to house executable scripts or programs that generate dynamic content in response to web requests.

handleConnection:
Isn't it rather the role of the parsing @dantol29 ? Anyway it should indeed be a separate issue. 


home.html :
I think it is good to have it as well as a separate issue, create various .html pages with nginx as a model

lmangall:
Actually I just remembered that the executable compiled from hello.c might sometimes need to have executables right `chmod + x`  so the compilation in the Makefile is not the way to go, the executables simply have to be present in the directory for programs.
On a fresh clone I didn't encounter any problems but still this has to be changed

lmangall:
- The "end functions" now return their answer instead of writting directly into a socket, wich is a first step into using poll()
- handleCGIRequest() now takes argv and envp as arguments
- I renamed functions and reorganized the files (html directory and individual files) to avoid merging conflicts and make collaboration easier

552020:
> 1. The source code for the script should be "kept away" ideally in a place not accesible from the web for security reasons (perhaps this is a perfect use for your secret directory)

I got the same advice from the friend, but in our case it should be the part of the project I would say. As far I understood is not common to have a CGI in C++, normally you have PHP or Python. Now I spent a couple of hours researching about the CGI. I'm still fuzzy about it but I gained much more clarity. 

This short article was also helpful: https://webdevelopmenthistory.com/1993-cgi-scripts-and-early-server-side-web-programming/

And also imagining the fact that CGI was 'invented' in 1993, where PHP and JavaScript only in 1995. One of the first languages used for CGI scripts was Perl but then PHP was invented for this very exact scope: makes web pages 'dynamic'. A normal web server (without CGI capabilities, back in 1993 or 1995) can just server static webpages (Wikipedia like wihtout editing and stuff). If we want to let our user to send subscribe to a mailing list, and we want to write the email address he or she communicates us into a database, we need a CGI script. If we want to display the current date (back in 1995) when the our webpage is loaded we need a Perl or PHP script for that. We will have a function in the backend that read the date in the system and copy/paste it in the right place. For example, this could be an index.php page that prints the date when the page is requested by a client. 

```php
<!DOCTYPE html>
<html>
<head>
    <title>Welcome to Our Homepage</title>
</head>
<body>
    <h1>Hello, welcome to our homepage!</h1>
    <p>Today's date is: <?php echo date('Y-m-d'); ?></p>
</body>
</html>
``` 
You probably already researched it but since I learned it 5 minutes ago I repeat it to reinforce my understanding. This index.php page is a PHP script. It's not a program in the sense our C and C++ programs are, cause it doesn't get compiled into an executable but it gets interpreted. There is a PHP interpreter (which is an executable, and that we would launch into execve when we want to build the response for the request of index.php. The interpretere will take a php file and transform it in a pure HTML (CSS/JS) file, substituting all the small <?php ... /> insertion with the proper values. PHP is basically never compiled into an executable, but it's just transformed into an html file, as string. The interpreter doesn't produce another file but a string. 

Coming back to the question about how hidden should be the 'source files' of these scripts. We can write also some sort of program written in C++ that if /index is requested gives us the same output than the index.php script of above but it's note done cause it's cumbersome. An PHP page looks like an html page, the page we are serving now with each line being printed out in C++ not. Our server in C++ will be compiled and on the server (the machine) we don't need the executable. We need the static content (the htmls pages) which will be not compiled, but read and served by the web server if they are requested and we could have .php or .pl (perl) files (which are not compiled but they will be like the .html files on our files system) or a C++ cgi executable that we could use to generate dinamic content. In this case there is no reason to have the source code on the server itself. But in general for our purpose if we want to have a CGI in C++ then we need the source code I would say in our repo for the same reason because we need the source code of our webserver in our repo. Otherwise the evaluators need to be neo to read our code. 

The model I guess was trying to say to just not to serve your scripts the same way you are servind stating content, i.e. don't copy your scripts into the reponses, the scripts needs to be executed and they should be kept apart from the static content which is served the way it is (and not run through an executable). 

> 2. g++ instead of c++ is probably a collateral change when I was trying to have the Makefile to compile the script source code
> 3. I think the CGI executed program is exempt of the limitation from the subject. It is not per se a part of our webserver and the work we have to deliver, only an addition to demostrate how it works.
>    In this case I thought the code could take a string query, and I modified the main so it passes an hardcoded string query through execve. I think our webserv should send even more environment variables to the CGI executed script. Actually I now think it would be better to have a php scrit instead of a compiled C++ binary, perhaps I will do that next

I got also a more clear idea of what we need to do about hthis. We just need to implement the CGI 'standard' or the CGI protocol as specified in the RFC 3875. https://www.rfc-editor.org/info/rfc3875. It seems that this is the first (and also the last one) standard for CGI. It has never updated, and soon probably other standards have been introduced. 

This standard just specifies how information should flow from any web server (if it wants to be CGI compliant) to any external CGI program that is used to generate dynamic content. For example, if we receive a http POST request for the cgi the body of the request should be passed through stdin (and not as env variable) but the rest of the information in the header as also explained in this snippets we had in MIro need to be passed through environment variables.  The subject forbids to use execve to run a web server but we can run the PHP interpreter with execve or something similar for python or for C++ (but we should use a sort of C++ web development framework like Crow). So yes, I would say we don't need to write our own CGI program that should be executed by execve. 

> In this case I thought the code could take a string query, and I modified the main so it passes an hardcoded string query through execve.

I dont' understand what you mean here. 

> 4. README_CGI
> 
> * now has a markdown formatted title
> * now explain the difference between programs and scripts (In a nutshell: scripts are interpreted at runtime, programs are compiled)
> * the good RFC: everything was ok, I removed this comment to myself
> 
> regarding the pdf: I took it out, it is already outdated, it was more intended as an internal (between us) note
> 
> Should the cgi-bin directory be hidden ? The Friend: By default, the cgi-bin directory is not "hidden" in the sense that web users can request and execute the CGI programs contained within it via their web browser. This is the directory's primary purpose: to house executable scripts or programs that generate dynamic content in response to web requests.
> 
Her my question was about .cgi-bin being hidden in the code but not in the README. 

> handleConnection: Stefano : _The handleConnection is a sort of router (before the cgi). We can write an extra issue for it and start to think about how it shoudl work cause I have no idea how to design it in the context of the server core functionalities._ -> Isn't it rather the role of the parsing @dantol29 ? Anyway it should indeed be a separate issue.
> 
Do you want to write these issue @lmangall? Also the one below or should I?

> home.html : I think it is good to have it as well as a separate issue, create various .html pages with nginx as a model

552020:
> Actually I just remembered that the executable compiled from hello.c might sometimes need to have executables right `chmod + x` so the compilation in the Makefile is not the way to go, the executables simply have to be present in the directory for programs. On a fresh clone I didn't encounter any problems but still this has to be changed

Don't we give exec rights only to scripts? I'm not sure about this at al.

552020:
> * The "end functions" now return their answer instead of writting directly into a socket, wich is a first step into using poll()

I didn't look at the code yet: but for my (still fuzzy) understanding this seems the way to go. The CGI module take the parsed http request (plus eventually the fd of the socket just to then be able to the 'core' where to sent back the reponse) and generates an http response. It just send the data in a CGI format to a CGI external program that will give back an http response as string, the CGI will not even need the parser to transform this string back into an http response and will give the string back to the core with the fd so that it can be send back to the client. Right?

> * handleCGIRequest() now takes argv and envp as arguments

argv in our webserver will be only the name of the server and eventually the configuration file. why do you need argv?

> * I renamed functions and reorganized the files (html directory and individual files) to avoid merging conflicts and make collaboration easier

Top!

lmangall:
>  if we receive a http POST request for the cgi the body of the request should be passed through stdin (and not as env variable) but the rest of the information in the header as also explained in this snippets we had in MIro need to be passed through environment variables

You are right I didn't do that. Currently the address of the script `./cgi-bin/hello_py.cgi` is hardcoded and passed as argv and I also have hardcoded an environment object that gets transformed into a string before being pass as env to execve (that runs the script).
Passing the body through stdin is particular to POST I thihnk.

So next I can try to:
1. identify the request
 2.pipe 
 3.pass body through stdin and metadata as env through execve()

lmangall:
Regarding opening issues to rewrite `handleConnection`
I think it is better, clearer to write the issue once this branch is pulled to main

> Don't we give exec rights only to scripts? I'm not sure about this at al.

Executables as well


> I didn't look at the code yet: but for my (still fuzzy) understanding this seems the way to go. The CGI module take the parsed http request (plus eventually the fd of the socket just to then be able to the 'core' where to sent back the reponse) and generates an http response.

At the moment `handleConnection` is writing a response(in the form of a string) to the socket, which it gets from `handleCGIRequest` `handleNotfFound` `handleHomePage` .
We don't have yet several clients, but I suppose you (as the core) will tell me wether you just want the answer, or want the answer and the fd (kindda client identification) back, so they "stay together"
PS: We have a scripts that hangs >30s



> It just send the data in a CGI format to a CGI external program that will give back an http response as string, the CGI will not even need the parser to transform this string back into an http response and will give the string back to the core with the fd so that it can be send back to the client. Right?

Right, the CGI produce ready to use http response



> argv in our webserver will be only the name of the server and eventually the configuration file. why do you need argv?

I am not talking about this "main" argv, but argv from execve (after the fork to execute the CGI) so I have to give argv as an argument, but indeed it doesn't serve any purpose.
It would be great to use it to pass the body of the request (rather than to pipe and stdin) but  it would  make the gods of the RFC3875 angry (and be bad practice) so... nope.

dantol29 (dismissed):
**Fantastic job!** I am sorry that I could not review it before. The program works, however I had to add a few changes to compile it. 
Here are the changes I had to make:

1. RequestHandler.cpp:14:71: error: no member named 'to_string' in namespace 'std'
                               std::string("Content-Length: ") + std::to_string(htmlContent.length()) + "\n\n" +
                                                                 ~~~~~^
1 error generated.

**I had to change -std=c++98 to -std=c++11, because to_string is from 11.**

2.CGIHandler.cpp:50:9: error: use of undeclared identifier 'waitpid'
        waitpid(pid, &status, 0);
        ^
1 error generated.

**I had to add #include <sys/wait.h>**

3. Environment.cpp:33:8: error: no member named 'strcpy' in namespace 'std'
                std::strcpy(envCStr, env.c_str());
 
 I **had to add #include <cstring>** 
 
 4. make re does not work
 error: unable to open output file 'obj/main.o': 'No such file or directory'
1 error generated.

lmangall:
Thanks, actually I started taking notes for our next meeting and I noted that being on ubuntu [you will be able to spot some problems ](https://miro.com/app/board/uXjVKbFlOrg=/?moveToWidget=3458764584528500239&cot=14)we can't if we are not using docker or a virtual machine. 
PS: feel free to add to the list of stuff to talk about on Miro

Regarding the changes, it should be the duty of the pusher to do them, correct @552020 ? But in some case it is good if we do them for somebody else, thank you

However this one :

> RequestHandler.cpp:14:71: error: no member named 'to_string' in namespace 'std'
> std::string("Content-Length: ") + std::to_string(htmlContent.length()) + "\n\n" +
> ~~~~~^
> 1 error generated.
> I had to change -std=c++98 to -std=c++11, because to_string is from 11.
> 

Is a pretty bad one, we can't push to main with this, we would fail or get -42 I will have to fix it.
It's here because on Mac I didn't get the error

dantol29:
Can you resolve the conflict? I merged Stefano's PR

CGIHandler.cpp
```diff
@@ -0,0 +1,63 @@
...
+        perror("fork failed");
+        exit(EXIT_FAILURE);
+    } else if (pid == 0) {
+        close(pipefd[0]);
+        dup2(pipefd[1], STDOUT_FILENO);
+        close(pipefd[1]);
+
+        if (!envp.empty()) {
+        // Convert the vector to a suitable format for execve
+            if (execve(argv[0], const_cast<char* const*>(argv), &envp[0]) == -1) {
```
552020, line 33: Do you understand this casting: `const_cast<char *const *>(argv)`. I did the module about casting, and I understand const_cast but the `char *const *` I don't understand. Instead of having a simple double pointer we have a double pointer and the second is const? Is it like this?

552020 (approved):
1. Only file names with classes (for my understanding) should be capitalized. If the file has only normal functions then the best practice should be have them lower case I guess. 

I open an issue for that so that it will not block this PR: https://linear.app/web-serv/issue/WEB-60/decide-about-capitalization-of-files-which-doesnt-contain-classes

2. I would keep for classes and the .hpp and the .cpp file together and not put them in an include directory but we can discuss how we can have. 

Same for this: https://linear.app/web-serv/issue/WEB-61/decide-about-splitting-headers-and-implementation-files-of-classes

3. If we have an include directory then we need also a src/ directory so that all the source files are not just in the root of the project. 

Same for this: https://linear.app/web-serv/issue/WEB-62/decide-about-having-a-src-directory-or-not

4. About the exit in the code, which is also a problem in the code I wrote I opened this issue: https://linear.app/web-serv/issue/WEB-33/decide-about-exits

5. About the Environment class, same as above: https://linear.app/web-serv/issue/WEB-64/decide-about-orthodox-canonical-form

6. There also some other small "issues" but all in all a good job, therefore:

LGTM!

552020:
@dantol29 needs to approve again!
@lmangall this branch has conflicts that must be resolved!

dantol29:
I will approve as soon as all the conflicts are going to be resolved

lmangall:
> std::toString is a common "issue". In MacOS is accepted in the C++98 standard in Linux is not. 

okay, that explains everything

________

> Template function is correct to have them in the .hpp file or in a .tpp file.

Thanks, I didn't knew about .tpp

_______

Regarding the casting, the man env has :
`int execve(const char *pathname, char *const _Nullable argv[], char *const _Nullable envp[]);`

so argv is a `char *const *` : a pointer to a constant pointer to characters, or as you wrote a double pointer and the second is const, it is the way it should be

________

> Only file names with classes (for my understanding) should be capitalized. If the file has only normal functions then the best practice should be have them lower case I guess.

I think we have some information about that in the introduction of the CPP modules, 

________

Thanks for opening all the issues !

lmangall:
I solved the conflicts, it resulted in this message:
Screenshot 2024-04-05 at 11 02 07

Therefore i re-request the review

552020 (approved):
LGTM

dantol29 (approved):
Now I can merge!

PR #31, opened 2024-03-30
implement so reuseaddr and so reuseport option for the listening socket

552020:
- super small optimisation that anyone has (see new doc about setsockopt, which will be used only once)
- plus rewritten memset (not allowed) but many have in their code
- plus replace printfs with std::cout


552020:
I agree to keep main as uncluttered as possible, so if someone suggest to move ft_memset in another directory I'll agree!

lmangall (approved):
Sorry, I missed this PR !

dantol29 (approved):
Small optimization approved!

PR #59, opened 2024-04-04
basic request-line parsing

dantol29:
Basic request-line parsing. It does not save variables in the Class, because we have not created it yet.

1. Checks for the pattern: method, SP, request-target, SP, protocol version.
2.  Makes sure there is CRLF at the end of the line.
3. Implemented MAX_URI(A server that receives a request-target longer than any URI it wishes to parse MUST respond with a 414 (URI Too Long) 
4. Implemented wrong method(A server that receives a method longer than any that it implements SHOULD respond with a 501 (Not Implemented) 
5. Checks if request-target exists with access()

Postman does not work on 42 PC, that is why I could not properly test it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ DEPFLAGS = -MMD -MP
 TARGET = webserv
 
 # Source and Object Files
-SRCS = main.cpp
+SRCS = main.cpp parseHttpRequest.cpp
 
 # Create an obj directory if it doesn't exist
 OBJDIR = obj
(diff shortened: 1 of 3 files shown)


dantol29:
image
You can see that it is linked to an issue here

552020 (commented):
We don't have any mean to test it. Right?

Should we write one or more unit tests around it?

I also don't understand why it seems that memset has been replaced in this branch even if in main we have already memset

image

dantol29:
Since the same code is implemented in the other branch, there is no point to merge it. I will close it.

PR #82, opened 2024-04-06
Lmangall/web 69 add file name rule to our internal style guide

lmangall:
`environment.cpp` but` Environment.hpp `
as by PascalCase for files containing classes I assume we are talking Class definition, correct ?

`CGI_handler.cpp` as CGI is an acronym

PR #80, opened 2024-04-06
Lmangall/web 70 add camelcase as the way to go for variable names in our codebase

lmangall:
I used` pipeFD` and `pid`, maybe you want to use rather `pipefd`...
https://linear.app/web-serv/issue/WEB-70/add-camelcase-as-the-way-to-go-for-variable-names-in-our-internal

PR #83, opened 2024-04-06
Lmangall/web 66 create a src folder

lmangall:
Please accept those PRs first:
https://github.com/dantol29/webserver/pull/82
https://github.com/dantol29/webserver/pull/80

Now we can work on the code with decided standards and well organized.


dantol29:
I thought we wanted to keep class.hpp and class.cpp in the same folder.

- Stefano @552020 
`A class will be separated in two files: MyClass.hpp and MyClass.cpp (both files will be in the same directory)`

lmangall:
indeed,` Environment.hpp HTTPRequest.hpp` are going to `src/`

552020 (changes requested):
Conflicts!

dantol29:
@552020 I thought HTTPRequst.cpp and .hpp go to /include

lmangall:
In the case of a class,
Do we want the .cpp in the include dir or the .hpp in the src dir ?
At the moment I did .hpp in the src dir...
I am fine with both

lmangall:
I checked a few peers code and they all have them separated: .hpp in include, .cpp in src, only Theo[ has a class folder in src](https://github.com/theozanchi/42_Berlin_webserv/tree/main/src), but he also separates

552020:
> In the case of a class, Do we want the .cpp in the include dir or the .hpp in the src dir ? At the moment I did .hpp in the src dir... I am fine with both

.cpp in include is not a viable solution (I guess) but .hpp wherever its .cpp peer is: .hpp in src is perfect! Thanks!

552020 (changes requested):
Could you add also src to the includes -I.src in the Makefile so that we can just write #include "MyClass.hpp"? Thanks!

lmangall:
Done, it should be ready to pull then

552020 (approved):
LGTM

dantol29 (approved):
Looks good

PR #84, opened 2024-04-06
53 parse origin form request target format

dantol29:
@lmangall Now parser stores all the variables. You can see how class is implemented. I will add more tests in the future


552020 (changes requested):
Conflicts with main!

552020 (dismissed):
LGTM!

If HTTPRequestUtils does not contain a class should be snake_case I guess.

dantol29:
It has the functions that the class is using.

dantol29:
### You can test how it works 

1. c++ tests.cpp -o test
2. ./webserver
3. ./test "query" or "simple"

dantol29:
@552020 @lmangall `/path/resource?query=string` when I receive the request with the path, does it mean that I should search for a resource in a `webserv/path/` and if no path provided in a root folder.

lmangall:
Basically yes, we are starting from root so "no path" means root, that's my understanding.
Will check the code tomorrow morning

lmangall:
You have some .cpp in the root of your branch, from merging a former version of main to your branch

lmangall:
Is the PR just for me to see or do you also want us to review and pull it ?

dantol29:
@lmangall Yes, I want you to review it because it has crucial functionality for CGI


I deleted the files that were in the root.

lmangall (approved):
Thanks !

lmangall:
THanks, yes I am definitely looking at your http class to implement my functions, currently related to the request target
https://github.com/dantol29/webserver/blob/lmangall/web-63-implement-cgi-request-meta-variables-into-the-variable/src/CGI_meta_variables.cpp

Just to clarify, "extract" is to extract from the request you get and "get" is to get from the object, correct ?

dantol29:
@lmangall Extract is used only inside of the class to parse the request and save the result in the attributes of the class. To get the attributes you have to use public member functions(get)

dantol29:
@552020 Can you review it again?

lmangall:
> Extract is used only inside of the class

I am not sure as you probably know better CPP than I do thanks to the modules, but I see that the extract functions (and a couple others) are declared outside of the class. shouldn't they rather be declared inside the scope of the class for encapsulation?

dantol29:
> > Extract is used only inside of the class
> 
> I am not sure as you probably know better CPP than I do thanks to the modules, but I see that the extract functions (and a couple others) are declared outside of the class. shouldn't they rather be declared inside the scope of the class for encapsulation?

As I know encapsulation is used to restrict access to the class attributes and define the way they are initialized or changed. These functions outside of the class do not have access to attributes and do not change them. That is why I think it is fine

lmangall:
I think what you are writing relates more to the private/public aspect, and that if the functions are "related" they shold be part of the class, because as much as possible everything has to be an object. But I am not sure at all, in my part one of the next things I have to change my code so my functions are part of an object. Perhaps @552020 will know more

552020 (commented):
Starting to review it now:

- Server.cpp, Utils.cpp, server.cpp are committed as empty files. Why is that?

- HTTPRequestUtils.cpp, if has general helper function (or also if it has helper function for a certain class) has to be snake_case. If af file is in PascalCase it means it has a class in it. It's not critical though. @lmangall what do you think?

dantol29:
> Starting to review it now:
> 
> * Server.cpp, Utils.cpp, server.cpp are committed as empty files. Why is that?
> * HTTPRequestUtils.cpp, if has general helper function (or also if it has helper function for a certain class) has to be snake_case. If af file is in PascalCase it means it has a class in it. It's not critical though. @lmangall what do you think?

Looks like when I was resolving conflicts empty files appeared.
I have already deleted them in the other PR

552020:
> @552020 @lmangall /path/resource?query=string when I receive the request with the path, does it mean that I should search for a resource in a webserv/path/ and if no path provided in a root folder.

I'm not sure if the HTTPRequest 'module'/object should search for anything. In my opinion it should blindly copy the request, in this case the path without looking if it exists or not. It will be the router, then, that will check if the resource exist (cause it is checking anyway for resources) and if it doesn't exist than it will set the proper error in the HTTPResponse object. 

If there is a path, the search should start, from the directory hosting the website, I guess it is specified in host, and this directory will be inside the directory set as the directory holding the content to be served (which will be a default value, like /var/www/ or the value given in the configuration file) from the subect: _Define a directory or a file from where the file should be searched (for example, if url /kapouet is rooted to /tmp/www, url /kapouet/pouic/toto/pouet is /tmp/www/pouic/toto/pouet)._

If I'm not wrong an http request doesn't have a status code, so an http request object should also not have a statu code I would say. 

I read the concern of Leo about these function being outside the object and I'm myself not sure why they are outside: can you justify this choice somehow?

dantol29:
HTTPRequest has the status code to simplify the work you have to do. If status code is not 200, you know something is wrong in the request and you can respond with an error. Otherwise, you will have to double check every case again, which does not make sense. 

HTTPRequestUtils exists just to split the code. HTTPRequest.cpp already has 214 lines and HTTPRequestUTils 289 lines. By the end of project they gonna have around 500 each. I really do not like files that have too many functions, it becomes unreadable at some point. I guess that is why we were allowed to have 5 function max in the file

552020:
Ok.

lmangall:
> HTTPRequestUtils.cpp, if has general helper function (or also if it has helper function for a certain class) has to be snake_case. If af file is in PascalCase it means it has a class in it. It's not critical though. @lmangall what do you think?

I agree, let's just do it in a next PR (that is not dedicated to this purpose I mean)

PR #94, opened 2024-04-08
CGI meta-variables (environment) from HTTPRequest object

lmangall:
_This code is not yet integrated into the broader codebase_

We have most of the RFC3875 meta-variables

> meta-variable-name = "AUTH_TYPE" | "CONTENT_LENGTH" |
>                      "CONTENT_TYPE" | "GATEWAY_INTERFACE" |
>                      "PATH_INFO" | "PATH_TRANSLATED" |
>                      "QUERY_STRING" | "REMOTE_ADDR" |
>                      "REMOTE_HOST" | "REMOTE_IDENT" |
>                      "REMOTE_USER" | "REQUEST_METHOD" |
>                      "SCRIPT_NAME" | "SERVER_NAME" |
>                      "SERVER_PORT" | "SERVER_PROTOCOL" |
>                      "SERVER_SOFTWARE" | scheme |
>                      protocol-var-name | extension-var-name

By "most" I mean that:
- The essential ones are ready to go 
- Some (REMOTE_HOST, REMOTE_ADDR, AUTH_TYPE, REMOTE_USER, REMOTE_IDENT) are just as draft as we may implement them, but don'thave to / wont

note from the RFC that I believe matters for the http response @552020 :

> The SERVER_SOFTWARE meta-variable MUST be set to the name and version
>    of the information server software making the CGI request (and
>    running the gateway).  It SHOULD be the same as the server
>    description reported to the client, if any.

__________________

note from the RFC @dantol29 regarding [our discussion on the last PR](https://github.com/dantol29/webserver/pull/88#issuecomment-2042767922):

>  If multiple header fields with the same field-name
   are received then the server MUST rewrite them as a single value
   having the same semantics. Similarly, a header field that spans
   multiple lines MUST be merged onto a single line.
   
   So for me it's fine if you concatenate several "values" into one as this is the final form


lmangall:
Only PATH_TRANSLATED is not implemented at the moment, I left it to the end as I was thinking maybe we'll change the structure again, but I might also do it tomorrow or before this PR gets pulled

dantol29:
I looked through the code, and everything totally makes sense. Only it doesn't make much sense to extract keys and value from the queryString to combine them again afterward. I can just save the raw queryString without  modifying anything and optimize the code a bit.

Overall, I like it a lot!

lmangall:
Thank you,
Regarding the query string it's true at the moment but I might need to concatenate it another way (I am not totally sure in what form I am supposed to pass it to the program/script) so at the moment I think we can keep it as is.

dantol29 (dismissed):
NICE

lmangall:
Next step could be issues WEB-87 and WEB-88

docs/CGI/README_CGI_CONCEPTS.md
552020, line 1: 1. Still not starting from # for the titles in Markdown. 
2. Do you think we need the README_CGI prefix, if the md documents are already in the CGI folder?
3. It would be interesting to follow the RFC (you added) to convert the HTTPRequest to another object like a HTTPCGIRequest

Otherwise good!

docs/CGI/README_CGI_ENV.md
552020, line 1: 1. The biginning is a little bit confusing visually. I would maybe avoid to just quote `>`

image

Maybe it would be useful to know what are meta variables used for. Are all environment variables also meta variables? Are meta variables called meta variables only in the context of the script execution (where they are environment variables) or more in the context of the CGI processing where they are not yet 'environment variables' of the child process?

2. What are all these other variables after the meta-variablen-name, like protocol-variable-name and so on?

3. Since the distinction between URI and URL is not perfectly clear to me: I got the an URL is a type or URI, can we have an example of URI which is not an URL?

4. ## How does a URI translates to a meta-variable: I

In this case we are using an URL as example, right?

5. we use a `std::string _requestTarget;` to map the request to CGI meta-variables:

Is this information relevant? Till now we had general information about CGI and meta variables, now we talk about our implementation. It feels strange, like a jump. 

6. ### Origin-Form
- **Approach**: For requests **directly to our server**, we extract the path and query from `_requestTarget`.
  - `SCRIPT_NAME`: Assign path component of `_requestTarget` for endpoint identification.
  - `QUERY_STRING`: Extract and assign query component from `_requestTarget` for additional data processing.
  
  This I don't understand, like what is Origin Form and also these other "Form" paragraphs? What is this about?
  
  7. It's not strictly related on what you did, but I keep on ask myself where did the POST and DELETE request will be elaborated. I guess in the CGI. Is thist right?
  
  Otherwise good!

docs/CGI/README_CGI_IMPLEMENTATION.md
552020, line 1: Why then don't add them in the description of a Linear Issue instead of having it as file here? Since more than docs is stuff to 'implement further' (the code I presume)?

src/Environment.cpp
552020, line 1: 1. ft_strcpy
2. Should the Doxygen comments in Classes be places in the Header or in the Impemetation file?
3. I guess I implemented a similar function in the Router which splits the path in directories and file name, just to understand that was completely useless, cause we can use stats with the full path, but maybe is useful for you. 
4. In this function void  `Environment::HTTPRequestToMetaVars(char* rawRequest, Environment& env)` are you parsing stuff by yourself directly from the http request string?

552020 (changes requested):
LGTM but I have some questions, which I wrote in separate comments to the single files.

docs/CGI/README_CGI_CONCEPTS.md
lmangall: > Still not starting from # for the titles in Markdown.

✅ I do everywhere, only the "implementation" readme is just notes (personal notes, so not formatted)
I wrote it in the code actually:
```
# Notes
### those are snippets from the mighty RFC 3875
### I will use them to implement further, then organize/markdown them in a better way

```

I also had an unclosed quote that messed up the formatting

_____________________________

> Do you think we need the README_CGI prefix, if the md documents are already in the CGI folder?

✅ You are right, this is an "heritage" from the previous form of the code
_____________________________

> It would be interesting to follow the RFC (you added) to convert the HTTPRequest to another object like a HTTPCGIRequest

❌ Not sure what you mean, is that in the RFC ? (perhaps I missed it)
At the moment I don't see the advantage but maybe when the complexity will increase...
____________________________

docs/CGI/README_CGI_ENV.md
lmangall: > The biginning is a little bit confusing visually. I would maybe avoid to just quote >
image

Perhaps because you looked only at the code, the Markdown represented is really clean I think:
image


✅ 
_____________________________


> Maybe it would be useful to know what are meta variables used for. 

❌ I decided not to put it in the documentation because I feel a bit that we are flooding in documentation, also it is quite easy to understand from the code and some other part of the documentation explains some meta-variables
_____________________________

> Are all environment variables also meta variables? Are meta variables called meta variables only in the context of the script execution (where they are environment variables) or more in the context of the CGI processing where they are not yet 'environment variables' of the child process?

✅  From my understanding and checking :
- Meta-variables are named parameters sent from the server to the script.
- Some of them are system environment variables, but not all, or not necessarily
- Even if they are not env vars, they are accessed as such (as instance via getenv())
=> added to the documentation
_____________________________


> What are all these other variables after the meta-variablen-name, like protocol-variable-name and so on?

✅ ❌ they are WIP:
https://linear.app/web-serv/issue/WEB-88/add-protocol-specific-meta-variables
_____________________________


> Since the distinction between URI and URL is not perfectly clear to me: I got the an URL is a type or URI, can we have an example of URI which is not an URL?

✅ it was not clear to me neither, I updated the documentation, now it is better
_____________________________


> ## How does a URI translates to a meta-variable:
In this case we are using an URL as example, right?

✅ "http://localhost:8080/myScript.cgi/view?id=123" is a URL, it provide information on where is the ressource
_____________________________


> we use a std::string _requestTarget; to map the request to CGI meta-variables:
Is this information relevant? Till now we had general information about CGI and meta variables, now we talk about our implementation. It feels strange, like a jump.

✅ I start the documentation as a way to understand, because Daniil had this var I wanted to transform it to a meta-var but then realized that it doesn't make much sense.
Perhaps this whole part of the documentation (## Request target to meta-variable) should go, what do you think ? 
_____________________________

> This I don't understand, like what is Origin Form and also these other "Form" paragraphs? What is this about?

✅ much of the answer to this is in the answer right above, and on [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3) , it is more relevant for the http request, maybe I shouldn't cover it at all
_____________________________


> It's not strictly related on what you did, but I keep on ask myself where did the POST and DELETE request will be elaborated. I guess in the CGI. Is thist right?
✅  F and J told me that not only in the CGI, but to be honnest I don't  see yet how it's possible for static content 
_____________________________

docs/CGI/README_CGI_IMPLEMENTATION.md
lmangall: It's kindda a personal note, it doesn't make sense to have it as a linear issue neither, at the moment it is snippets of things I wanted to save to come back to later, but I can erase them if they disturb you (like, really, I understand that it looks dirty)

src/Environment.cpp
lmangall: > ft_strcpy

✅ Thanks for spotting, I got rid of it

lmangall: > Should the Doxygen comments in Classes be places in the Header or in the Impemetation file?

In C I put them in the header but here it seems too messy to me to put it in the Header, maybe it is possible to put them in the header below the object with a second declaration ?

=> The friend suggest the Header, I also learned that there are "doxygen comments"
` MyClass(); ///< Constructs a new MyClass object.`

=> answer from the friend

> Yes, it's possible to put the detailed Doxygen comments below the class declaration in the header file, along with a second declaration of the function. This way, you can keep the function definition in a separate source file (.cpp file), while still providing detailed documentation in the header. Here's how you can do it:

 ✅ seems like a solution, I think I will do that on my ride home today

`// MyClass.h

#ifndef MYCLASS_H
#define MYCLASS_H

class MyClass {
public:
    MyClass(); ///< Constructs a new MyClass object.
    ~MyClass(); ///< Destroys the MyClass object.
    int getData() const; ///< Gets the value of the data member.
    void setData(int value); ///< Sets the value of the data member.

private:
    int data; ///< The data member of the class.
};

/**
 * @brief Detailed documentation for MyClass constructor.
 * 
 * This constructor initializes the data member to a default value.
 */
MyClass::MyClass(); 

/**
 * @brief Detailed documentation for MyClass destructor.
 * 
 * This destructor cleans up any resources held by the object.
 */
MyClass::~MyClass(); 

/**
 * @brief Detailed documentation for getData member function.
 * 
 * @return The value of the data member.
 */
int MyClass::getData() const; 

/**
 * @brief Detailed documentation for setData member function.
 * 
 * @param value The new value to set.
 */
void MyClass::setData(int value); 

#endif // MYCLASS_H
`

lmangall: > I guess I implemented a similar function in the Router which splits the path in directories and file name, just to understand that was completely useless, cause we can use stats with the full path, but maybe is useful for you.

_stat() function: It retrieves information about a file specified by its filename and fills a struct stat with the file's attributes._

Thanks I didn't knew stat, I will have to look at that some more

lmangall: > In this function void Environment::HTTPRequestToMetaVars(char* rawRequest, Environment& env) are you parsing stuff by yourself directly from the http request string?

No, from the `std::string _requestTarget;` variable

552020 (approved):
LGTM!

552020 (approved):

dantol29 (approved):
You accidentally modified HTTPRequest.hpp, but it is not a big deal.
Besides that looks good!

PR #102, opened 2024-04-09
Parse body and more tests

dantol29:
I added parseBody member function that stores each line of a body in the `std::vector<std::string> _body`.

Also I made more tests for headers and body in the `tests.cpp` file


552020 (approved):
LGTM: nice!

I didn't understand how the test work. compiled the test.cpp file and I run it like ./test "simple" is this correct. But then nothing happen. 

After I'm done (hopefully soon) with my branch we should compare our work on the chunked bodies cause I also wrote some logic for that. 

I have also a class HTTPResponse with a method setStatusCode, to set the statusCode on the response object ...

dantol29:
Hmm, it should work.
@552020 
Did you launch webrowser before launching the tests?

552020:
> Hmm, it should work. @552020 Did you launch webrowser before launching the tests?

🙈 It works sorry! I didn't launch the server before I was kind of expecting something coming out from the terminal where I run test ...

lmangall (approved):
I did some tests and it all seems to working fine

PR #113, opened 2024-04-11
Parse configuration file

dantol29:
### Grammar I think is suitable
```
server {
	[TAB][KEY][SP][VALUE][;] // variables

	location /path {
		[TAB][TAB][KEY][SP][VALUE][;] // location variables
	}
}
```
### Grammar rules (that we have to define)
1. Empty lines are allowed only inside of the server {} scope. Outside of server {} scope, empty lines are disallowed.
2. Keys **MUST** not contain spaces.
3. Values **MAY** contain spaces.
4. Key-value pairs **MUST** be terminated by a semicolon ';'.
5. Every scope **MUST** be closed.
6. _maybe comments, but seems too complicated_
7.  And so on ...
#### 1. New `ConfigFile.hpp` class to get data
-  to initialize just pass path to the `ConfigFile(path)` constructor
-  all the variables are stored here `std::map<std::string, std::string> _variables;`
you can get them with `std::map<std::string, std::string> getVariables()` or `std::pair<std::string, std::string> getVariables(std::string name)`
- all the "location" variables are stored here `std::vector<std::map<std::string, std::string> > _locations;`
you can get them with `std::vector<std::map<std::string, std::string> > getLocations();`
- you can get an error message with `std::string getErrorMessage();`. If it returns an empty string it means that config file is valid

### Questions
- Do I have to check each key-value pair? If yes, then do we have a list of valid variable names?
- What do you think about this coding style `if (line == NULL) { break ; }`? Is it readable for you? I wrote some code using this type of if-else


dantol29:
## I added .sh file in the tests folder

What do you think? Is it the right format for the tester?

lmangall (changes requested):
As we discussed in person:
-please change the name of	`parseLocation(char *line, int fd);` and/or `parseLocationLine` to something clearer

I would approve your pr just with the minimal, above change but it would also be good to confirm wether GNL is the best choice for this (maybe there is a stream function that is more appropriate)

Otherwise great code, it's good that you added the grammar as a comment upwards the function

lmangall:
Sorry I forgot to add a comment about your questions

> Do I have to check each key-value pair? If yes, then do we have a list of valid variable names?

Yes I think in order to accept only valid content, there is not a list available I think, we should do one, with the minimal requirements and additionnal ones 


> What do you think about this coding style if (line == NULL) { break ; }? Is it readable for you? I wrote some code using this type of if-else
The thing is that it would means not only changing the if in this case, but it could be changing the if formatting in many cases, [when we'll all be on the same page with the autoformatter](https://linear.app/web-serv/issue/WEB-100/check-if-everyone-is-up-to-date-with-the-autoformatter) we can decide (for me it's fine). It might be a goog exercise to manipulate the autoformatter to add your preferences...

552020:
It looks good, thanks! A

1. A question for my understanding: are these grammar rules, something defined in somen RFC?

Regarding your questions:

> Do I have to check each key-value pair? If yes, then do we have a list of valid variable names? 

Which key-value pairs? A list of which variable names? 
(Maybe also @lmangall can explain this, since he alrady answered the question.)

> What do you think about this coding style if (line == NULL) { break ; }? Is it readable for you? I wrote some code using this type of if-else

For me is ok, as far as there is an option in clang-format to have short if statements formatted like this, and we all have the same ...

dantol29:
@lmangall I changed functions. Now it is saveVarible instead of parseLine and saveLocationVarible instead of parseLocationLine.

lmangall (dismissed):
Good job!

dantol29:
# Parser checks the values of the variables

## Valid variables values
- listen [port_number]
- host [???]
- server_name [example.com] [www.example.com]
- :ballot_box_with_check: error_page [500] [502] [503] [504] [/custom_50x.html]
- index [index.html] [index.htm] [index.php;] // will first look for index.html. If that file isn't found, it will look for index.htm, and if that isn't found ...
- root [/var/www/example.com/html]
- :ballot_box_with_check: client_max_body_size [20000]
- :ballot_box_with_check: autoindex [on]
- :ballot_box_with_check: allow_methods [GET] [POST]
- alias [/var/www/documents/]
- cgi_path [/usr/bin/python3]
- :ballot_box_with_check: cgi_ext [.py] [.pl]

552020:
After the changes in the main branch, there are some conflicts that need to be resolved ...

dantol29:
@lmangall @552020 I added operator<< overload, if you wanna print all the values from HTTPRequest or ConfigFile just `std::cout << obj << std::endl`

listen [port_number] @552020 Any number?
host [???] @552020 IDK
server_name [example.com] [www.example.com] @552020 Anything?
☑️ error_page [500] [502] [503] [504] [/custom_50x.html]
☑️ index [index.html] [index.htm] [index.php;] // will first look for index.html. If that file isn't found, it will look for index.htm, and if that isn't found ...
☑️ root [/var/www/example.com/html]
☑️ client_max_body_size [20000]
☑️ autoindex [on]
☑️ allow_methods [GET] [POST]
☑️ alias [/var/www/documents/]
☑️ cgi_path [/usr/bin/python3]
☑️ cgi_ext [.py] [.pl]

lmangall (dismissed):
We can find a better name than "a" for the configfile but otherwise it looks great ;)

config
```diff
@@ -0,0 +1,26 @@
+server {
+	listen 8081;
+	host 127.0.0.1;
+	server_name free_money_near_u;
```
552020, line 4: Is this part of a plan or just a random name? :-)

config
```diff
@@ -0,0 +1,26 @@
...
+	allow_methods GET POST DELETE;
+	autoindex on;
+	cgi_ext .py .php .pl;
```
552020, line 7: regarding the `cgi_ext` do we need here also .cgi? If you (@lmangall) want to have also CGI scripts which are programs (like .c/.cpp cgi files) we need a way to distinguish them from the scripts (the .php files for examples) which are interpreted, right? The .c/.cpp file would be passed as first argument of argv to execv and the second ones as second arguments being the interpreter the first one.

lmangall: Yes I already asked @dantol29 in person to add a .cgi
 
 argv[0] will be the script or program itself, as an executable, so .c or .cpp but **compiled**
 argv[1 and more] would **rather** be a file taken by our program
 
A query string will not be passed to a CGI script/program in **argv, it will be passed as a meta-variable (called environment variable once it is not in the CGI anymore but in the program).

lmangall:
Surprisingly difficult to find info on port number
under 1024 is most likely reserved, over is OK, 8080 is standard

IANA seems to be an authority

This document refers to several RFC:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml
There is a list here, but it's the kind of list you already need to know what you want to find
https://svn.nmap.org/nmap/nmap-services

config
dantol29: just a random name, it does not matter at all

config
552020: This means is correct not to list here (cgi_ext) the .c and .cpp extensions even if we will server pages generated through .c/.cpp scripts. The distinction on the different way to call the CGI handler will be 'hardcoded' in the Router and it will not be something that can be changed anyway in the config file or should we have also a settings for the file extensions of compiled languages files which can run a script directly (without a taking a file as argument), like you could have .c or .cpp or both or none?

dantol29:
Can someone mess around with the config file a bit to check the parsing? Maybe I did not cover everything...
It is easy and straightforward, just modify config and see how webserver reacts

lmangall:
As we  discussed in person:


Hackers will try absolute paths and `../`
```
	location / {
		root ../Documents;
	}
```
```

	location / {
		root ~/;
	}

```
```

	location / {
		root /usr/bin;
	}
	
```
```

	location /path {
		../../index;
		cgi_ext .py .cgi;
	}

```

__________________________________

Probably reserved until 1024:
`	listen 0001;
`
__________________________________


check that user doesn't ask for a reserved IP

lmangall:
@apparently this is called "Path traversal vulnerabilities" AKA "Carlo's candy"

dantol29:
Wow, that is interesting....
Sadly, we will have to upset Carlo, I just fixed it ...
```
bool	isVulnerablePath(const std::string& path)
{
	if (path[0] == '/')
		return (true);
	if (path[0] == '.' && path[1] == '.' && path[2] == '/')
		return (true);
	if (path[0] == '~' && path[1] == '/')
		return (true);
	return (false);
}
```
However, it is implemented for the configuration file. It has to implemented for requestTarget too

lmangall (approved):
Good job, thank you for making cleaning up the utils from your objects by encapsulating them into the class

lmangall:
@dantol29  Is there a reason for putting the operator overload (to print) out of the function ?

dantol29:
Yes, there is a reason. Left operator (std::cout) must remain unmodified. Otherwise, you would return something and modify it.

lmangall:
Okay thank you, I implemented two for the Environment and Response as function member (friend type) because some private vars needed to be accessed, I will read more about it and eventually change

552020:
> As we discussed in person:
> 
> Hackers will try absolute paths and `../`
> 
> ```
> 	location / {
> 		root ../Documents;
> 	}
> ```
> 
> ```
> 
> 	location / {
> 		root ~/;
> 	}
> ```
> 
> ```
> 
> 	location / {
> 		root /usr/bin;
> 	}
> 	
> ```
> 
> ```
> 
> 	location /path {
> 		../../index;
> 		cgi_ext .py .cgi;
> 	}
> ```
> 
> Probably reserved until 1024: ` listen 0001; `
> 
> check that user doesn't ask for a reserved IP

Yes, already read about `/../` the Router should just delete them or return an error. But I was thinking of this as a problem in the URL not in the config file of a server. Is the person going to use the server hacking himself? This doesn't make that much sense to me. But we can add an analogous check on the config file, where something like `/../` is just not allowed.

dantol29:
@552020 Can you merge this branch since I fixed that vulnerability

552020:
Sorry for the delay but I didn't manage to fully read it. Is it blocking your further work? But it's on my task list...

dantol29:
@552020 HTTPRequest class is changed a lot here that is why I want it to be merged ASAP

552020:
Ok. Then I'll just approve it.

552020 (approved):
LGTM!

PR #129, opened 2024-04-13
let code compile on my linux

552020:
some changes. since it's not a PR to main the double check is not needed (not even one) to merge, but I'll wait for your approval of course!


lmangall (approved):
Thank you !

PR #130, opened 2024-04-13
File upload(parsing part)

dantol29:
### Parser looks for the "content-type: multipart/form-data; boundary=..." and extracts boundary.
@552020 However, I cannot test it, because when I run this 
```
/client $'POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-type: multipart/form-data; boundary=stop_please\r\nContent-length: 42\r\n\r\nblablabla\r\n\r\n'
```
I get `poll client socket error: Bad file descriptor`

_./client is from docs/daniil_socket_


dantol29:
@552020 When our server receives a file upload HTTPRequest extracts boundary, but does it have to parse the body in a different way? I can implement it, but if we do not discuss, it gonna be the same story as with the chunked request.

552020:
> @552020 When our server receives a file upload HTTPRequest extracts boundary, but does it have to parse the body in a different way? I can implement it, but if we do not discuss, it gonna be the same story as with the chunked request.

Basically for the core doesn't make any difference if it is a file or not (if it's a `Content-type: multipart/form-data` or not). The only difference if it's chunked or not. If it's chunked then it will proceed with the chunked procedure, if it's not then not. 

The boundaries are extracted by the parser. The 'core' is interested only in the body. The boundaries are part of the body and the core will never know of them or parse them.

552020:
> ## Upload boundary is saved in the HTTPRequest
> tested and added get method.
> 
> @552020 Should I do anything else from my side regarding file upload?

I did a little bit of research and I  created a linear issue for your part of the implementation: https://linear.app/web-serv/issue/WEB-124/implement-parser-side

dantol29:
## Parsing is ready (it was surprisingly hard)

_Shout-out to Stefano for giving me a clear idea how to parse and for the suggestion to use structures_ 

@lmangall I guess it is useful information for the router

1. Each file is stored in a structure
```
struct File
{
    std::map<std::string, std::string> headers;
    std::vector<std::string> fileContent;
};
```
2. All structures are stored here
```
std::vector<File> _files;
```

## I NEED HELP!!!! @552020 @lmangall 
This member function does not work...
```
std::vector<File> HTTPRequest::getFiles() const
{
	return (_files);
}
```

dantol29:
## This is how it looks!
image

552020:
@dantol29 Is help still needed? Or could you solve it? What is not working in that function? Can you eventually open a linear issue?

dantol29:
@552020 I could not solve it...
It complains about the return type
```
src/HTTPRequest.cpp:151:9: error: no viable conversion from returned value of type 'const vector<HTTPRequest::File>' to function return type 'vector<File>'
        return (_files);
               ^~~~~~~~
```

lmangall:
I also  tried a few things and couldn't solve it,

dantol29:
## I added file upload to index.html to test

Currently, it works. The only thing is that headers are sent in the weird format, and I do not save them correctly. 
`Header1: value1; name="text"`

552020:
> ## I added file upload to index.html to test
> Currently, it works. The only thing is that headers are sent in the weird format, and I do not save them correctly. `Header1: value1; name="text"`

Is then still in "Draft". Or is ready to be merged? You mean weird format cause they are not divided by a new line like in the normal header of the http request? We need then first to split on ";" and then on "=", right?

dantol29:
## I changed response.setStatusCode(400) to setStatusCode(400, std::string message);

It just prints it in the terminal. It helps a lot to spot where the error comes from and simplifies Parses code a lot

dantol29:
## Now Parser is working correctly

_tested with tests.cpp_

dantol29:
## I cannot test it

We have to handle or at least allow POST requests

552020:
> We have to handle or at least allow POST requests

This should have been @lmangall part, but not he's gone, so maybe or we implement it or we need to test it with a sort of unit test.

lmangall:
This code and struct is not in main yet, maybe we can first merge it, if it is ready to be, and then work on the parser part (maybe I'll have time to do a mvp for it), what do yout think @dantol29  @552020   ?

552020:
I'm closing this one cause @dantol29 said it has been deprecated in favor of #196

PR #134, opened 2024-04-13
remove fd from vectors when socket are closed

552020:
A first try. Can you check @dantol29 if the problem persist? And can you give me the request that trigger the bad file descriptor thing.

### DIFF
--- a/docs/daniil_socket/Makefile
+++ b/docs/daniil_socket/Makefile
@@ -1,0 +1,24 @@
+CC=g++
+CFLAGS=-Wall -std=c++98
+
+all: client server
+
+client: client.o
+	$(CC) $(CFLAGS) -o client client.o
+
+server: server.o
+	$(CC) $(CFLAGS) -o server server.o
+
+client.o: client.cpp
+	$(CC) $(CFLAGS) -c client.cpp
+
+server.o: server.cpp
+	$(CC) $(CFLAGS) -c server.cpp
+
+clean:
+	rm -f *.o client server
+
+fclean: clean
+	rm -f client server
+
+re: fclean all
--- a/docs/daniil_socket/client.cpp
+++ b/docs/daniil_socket/client.cpp
@@ -6,24 +6,53 @@
 #include <iomanip>	// std::hex, std::setw, std::setfill
 #include <string>
 
-// Helper function to print characters in hexadecimal
 void printHex(const std::string &data)
 {
 	std::cout << "Hexadecimal representation:" << std::endl;
-	// in C++98 without range-based for loop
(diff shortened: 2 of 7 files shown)


dantol29:
```
./client $'POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-type: multipart/form-data; boundary=stop_please\r\nContent-length: 42\r\n\r\nblablabla\r\n\r\n'
```
@552020 This is the test request. 

./client is located at docs/daniil_socket

552020:
Did you check if the problem persist (on this branch)? @dantol29 

Cause I tested it now with the request you posted above but I didn't get the bad file descriptor error

image

dantol29:
OK, I will test ASAP.
Would be nice to put all those comments in a operator<< overload of it is possible in this case.

552020:
> Would be nice to put all those comments in a operator<< overload of it is possible in this case.

Sure, but what do you mean? Ok. I guess I understand what you mean. Yes we can do this but it will work only for objects, like when we print an object.

dantol29:
@552020 Why when I run that test request I do not see `std::cout << "Received HTTP request: " << std::endl << httpRequestString << std::endl;` this part in the terminal printed out?

552020:
> @552020 Why when I run that test request I do not see `std::cout << "Received HTTP request: " << std::endl << httpRequestString << std::endl;` this part in the terminal printed out?

I'll investigate.

552020:
Can you check again if it is working (check also if your request is well formed, i.e. if the declared body size in the header corresopnd to the size of the body)? And could you help me to solve this issue: https://linear.app/web-serv/issue/WEB-122/help-stefano-to-fix-content-length-retrieval-feature-in-daniils-client!

dantol29:
image
It still does not work

```
./client $'POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-type: multipart/form-data; boundary=stop_please\r\nContent-length: 3\r\n\r\nbla\r\n\r\n' 
```

552020:
Did you pull the branch? Cause it shoud print more information ...

I also open a PR regarding the parsing discussion in the core: https://linear.app/web-serv/issue/WEB-123/decide-if-the-core-has-to-do-any-parsing-at-all-or-not

dantol29:
_Of course I did not pull!_

## NOW IT WORKS! thanks!

lets merge ASAP!

dantol29 (approved):
WORKS!

lmangall (approved):
Well done, I know you guys checked this enough so I just reviewed the files and don't make you wait for my review.

PR #117, opened 2024-04-11
Lmangall/web 99 have cgi module with meta variables running with the current

lmangall:
MVP of the meta-variables of the CGI and of the new class `CGIHandler : public ARequestHandler ` object, you can access the CGI with `http://localhost:8080/hello` , the path is hardcoded at the moment but will rely on the meta-variables (that will rely on the config file) soon.
@552020  in order to keep your draft of the CGI object, which was nice with inheritance, I had to take several files from your branch currently in development

I started implementing in the branch `lmangall-temp` and then tried rebasing the commits from there but then gave p and copied the files, so some commits are there (modification to the environment object and creation of the CGI object)

PR #147, opened 2024-04-15
Lmangall/web 116 have router directing traffic

lmangall:
Building up on @552020 's work, Router is now routing, you can as instance check the following pages:

http://localhost:8080/python_site/cgi-bin/index.py
http://localhost:8080/perl_site/fun_fact/php_not_perl.html
http://localhost:8080/python_site/index.html
http://localhost:8080/hello_py.cgi
http://localhost:8080/hello_var.cgi

Ideally [PR 99 have cgi module with meta variables running ](https://github.com/dantol29/webserver/pull/117) PR should be merged first


dantol29 (approved):
## I got 1 errror
```
src/Server.cpp:66:39: error: too few arguments to function call, expected 2, have 1
                                                handleConnection(_connections[i]);
                  
```
Would be good to implement parser error codes and error messages in the router, because why would you send the request further if the parser says it is wrong.

## Overall, great!

552020 (changes requested):
LGTM but has conflicts.

dantol29 (dismissed):
HTTPRequest does not exist anymore, but LGTM!

552020:
> HTTPRequest does not exist anymore, but LGTM!

What does it mean that HTTPRequest doesn't exist anymore? @lmangall @dantol29

dantol29:
Sorry, I meant HTTPRequestUtils does not exist. It is moved as a private member to the class itself

lmangall:
Ok and you mean it doesn't exist "in main" I thought you were saying I deleted something. Thank you I will delete in my branch

lmangall:
@552020  Ready to be merged I believe, all the issues from #117  have bee addressed, some of the important ones:
https://github.com/dantol29/webserver/pull/117#discussion_r1569177786
https://github.com/dantol29/webserver/pull/117#discussion_r1569224269

Router is now better handled, however it is a part of handleConnection (a few lines inside it), eventually if you would prefer I can put everything in routeRequest and have it return a responseString ready to write to the FD...

552020:
> @552020 Ready to be merged I believe, all the issues from #117 have bee addressed, some of the important ones: [#117 (comment)](https://github.com/dantol29/webserver/pull/117#discussion_r1569177786) [#117 (comment)](https://github.com/dantol29/webserver/pull/117#discussion_r1569224269)
> 
> Router is now better handled, however it is a part of handleConnection (a few lines inside it), eventually if you would prefer I can put everything in routeRequest and have it return a responseString ready to write to the FD...

Thank you! The Router like everything happening as elaboration and response to a http request, given the actual design, should happen in the handleConnection method of the Connection object. So that's the right place! Few lines are fine. I should also outsource all these reading function but their destiny is not clear yet.

include/webserv.hpp
```diff
@@ -2,12 +2,20 @@
...
 
 #include <string>
-#include "Environment.hpp"
 #include <sstream>
 #include <cstdlib>
 #include <string.h>
+#include "MetaVariables.hpp"
 #include "HTTPRequest.hpp"
 
+// #include "MetaVariables.hpp"
```
552020, line 11: You wanted to remove all the commented out stuff right?

include/webserv.hpp
```diff
@@ -37,23 +45,7 @@ std::string readHtml(const std::string &filePath);
...
- * response is generated instead.
- *
- * @param argv An array of char pointers representing the arguments to be passed to the CGI script,
- *             including the script path as the first argument.
- * @param env An Environment object containing the environment variables to be passed to the CGI script.
- * @return A string representing the HTTP response generated from the CGI script's output, or a 500 Internal
- *         Server Error message if the script execution fails.
- */
-// std::string handleCGIRequest(const char *argv[], Environment env);
+std::string handleHomePage();
```
552020, line 48: Should this not be a method of the static response handler or be removed if not used?

src/CGIHandler.cpp
```diff
@@ -1,8 +1,6 @@
 #include "CGIHandler.hpp"
-#include <vector>
-#include <sys/types.h>
-#include <sys/wait.h>
-#include <unistd.h>
+
+#define EXIT_FAILURE 1
```
552020, line 3: This define is still there.

include/webserv.hpp
lmangall: Must have got mixed up with the two branches - gone

src/CGIHandler.cpp
lmangall: Sorry, must have got mixed up with the two branches - gone

include/webserv.hpp
lmangall: I think this can have its linear issue, it is in a way related to the way we serve errors, and to wether we are expecting an index.html in a directory :
https://linear.app/web-serv/issue/WEB-135/decide-of-a-custom-error-for-isdirectory-andand-havenoindexhtml

src/CGIHandler.cpp
```diff
@@ -12,10 +10,145 @@ CGIHandler::~CGIHandler()
...
+		argv[1] = NULL;
+	}
+
+	return argv;
+}
+
+HTTPResponse CGIHandler::CGIStringToResponse(const std::string &cgiOutput)
+{
 	HTTPResponse response;
-	// Determine CGI script path and set up environment
```
552020, line 19: I ask myself if is conceptually correct to define the HTTPResponse hier. Theoretically the response should be declared on the top of the handleConnection function and I guess it would be good if only one response object would go through the whole process and not to create different response objects in different places. Cause already the HTTPRequest/Parser could "fill" an HTTPResponse with an error code (which would by way bettter than set an error code on HTTPRequest object, which should not have a status code). But this is more something we can think about than a change request.

include/webserv.hpp
lmangall: Done: https://linear.app/web-serv/issue/WEB-136/decide-on-the-index-of-our-webserv

src/CGIHandler.cpp
```diff
@@ -12,10 +10,145 @@ CGIHandler::~CGIHandler()
...
+
+	std::size_t headerEndPos = cgiOutput.find("\r\n\r\n");
```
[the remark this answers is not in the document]
lmangall: DRY Alarm is correct, this is the thing I explain here: https://github.com/dantol29/webserver/pull/117#discussion_r1569177786
It allows the code to be clean by separating header and body returned by the CGI

src/CGI_handler.cpp
```diff
@@ -4,55 +4,65 @@
```
552020, line 4: Is this file really still here?

lmangall: I guess this one and server.cpp with a small "s" are here to doom us, this time you caught this one before it made its way into main!

src/Connection.cpp
```diff
@@ -366,6 +367,23 @@ bool Connection::readBody()
 	return true;
 }
 
+bool Connection::readRequestHeadersAndBody()
```
552020, line 370: What is that?

src/CGIHandler.cpp
lmangall: I see what you mean but indeed we have to thing about it, this way also allows us to have a method that returns the response, at the moment it made more sense than to have a method "filling" some of the attribute, but maybe it is not the best.

Actually we should have a red print if we write an error code on top of an already existing error code, I feel this would help a lot for debugging

src/HTTPResponse.hpp
```diff
@@ -20,11 +20,18 @@ class HTTPResponse
...
 	const std::string &getStatusMessage() const;
 	void setStatusCode(int statusCode);
+	std::string getBody() const;
+	std::string getHeader(const std::string &name) const;
 
 	void setHeader(const std::string &name, const std::string &value);
 	void setBody(const std::string &body);
 	std::string toString() const;
 
+	bool isCGI() const;
```
552020, line 30: A response should be just a response. Just an HTTPResponse. The responseHandler could be eventually the instance interested in knowing if a response it is serving is a dynamic one or not. But an HTTPResponse object should not carry this information. Or how can you justify it?

src/MetaVariables.cpp
```diff
@@ -0,0 +1,284 @@
...
+	{
+		if (str[i] != prefix[i])
+		{
+			return false;
+		}
+	}
+	return true;
+}
+
+// bool MetaVariables::isAuthorityForm(const HTTPRequest &request)
```
552020, line 108: Do you need all this commented out part?

src/Connection.cpp
lmangall: Was meant to declutter handleConnection, at the beggining when I took over the router, you don't like it ?

src/Router.cpp
```diff
@@ -12,13 +12,29 @@ Router::~Router()
 HTTPResponse Router::routeRequest(const HTTPRequest &request)
 {
 	HTTPResponse response;
```
552020, line 14: A response is created also in the Router.

src/HTTPResponse.hpp
lmangall: > A response should be just a response

Yes and being CGI is precisely part of being a response...

Being CGI, or not being CGI, is quintessential for a response, it is the essence of the response (not kidding) so I think it belongs here yes

src/Router.cpp
```diff
@@ -12,13 +12,29 @@ Router::~Router()
...
+			std::cout << "Path does not exist" << std::endl;
+			StaticContentHandler staticHandler;
+			response = staticHandler.handleNotFound();
+		}
+		else
+		{
+			StaticContentHandler staticContentInstance;
+			response = staticContentInstance.handleRequest(request);
+		}
 	}
```
552020, line 38: Can we create the CGIHandler and the StaticContentHanlder the same way? Like named or not named but the same way? I'm refering to line 20 and line 35/36

src/MetaVariables.cpp
lmangall: https://github.com/dantol29/webserver/pull/117/files/19b7efdf5c56ab1f0617af11a3ffb949d4f90b95#r1569252302

src/Router.cpp
lmangall: Sure, we can

src/Server.cpp
```diff
@@ -129,71 +128,17 @@ void Server::handleConnection(Connection conn, size_t &i)
...
-	// const char* argv[] = { "./cgi-bin/thirty_py.cgi", NULL };
-	// const char* argv[] = { "./cgi-bin/hello.cgi", NULL };
+	// NOTE: end of buffering/parsing part, start of router
 
-	// std::string response;
 	Router router;
+	std::string httpRequestString = conn.getHeaders() + conn.getBody();
+	HTTPRequest request(httpRequestString.c_str());
+	std::string responseString;
 	HTTPResponse response;
```
552020, line 138: a response object is declared here as well.

src/Router.cpp
lmangall: I am not sure if we are creating an object, I think at this level the model might be wrong:
https://chat.openai.com/share/2b9c431e-827a-48d5-821f-0395b7f2b022

But it says it doesn't matter, so it seems to be just a matter of what is aesthetic and easy to read, what do you think ?

src/utils.cpp
```diff
@@ -24,22 +24,22 @@ std::string readHtml(const std::string &filePath)
 	return buffer.str();
 }
 
-char *ft_strcpy(char *dest, const char *src)
-{
-	int i;
+char *ft_strcpy(char *dest, const char *src) {
```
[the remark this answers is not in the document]
lmangall: `namespace "std" has no member "strcpy"C/C++(135)`
This was the cpp 98 difference between mac and linux, we don't have it on school computers

src/Server.cpp
lmangall: answered [here](https://github.com/dantol29/webserver/pull/147#discussion_r1569447552)

src/Router.cpp
lmangall: But I agree it is better to create or declared the object only once, I will modify the function, except maybe those from CGIHandler

552020: I was more saying that we declare HTTPResponse like `HTTPResponse response` in many places in the code and actually we should create just 1 HTTPResponse object for each Connection. But if it works probably at the moment it doesn't matter. They will be just extra response object that get just destroyed or not created.

src/CGIHandler.cpp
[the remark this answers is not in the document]
552020: I see!

src/utils.cpp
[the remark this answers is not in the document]
lmangall: It's like that on my side, but yes we talked about it:
image

src/Router.cpp
lmangall: Yes that's what I understood,  they might be " temporary objects". Anyway I think you are right and I got rid of all of them I believe. it seems a bit harder to read but it is probably because of C vs C++ and it seems to be good practice

src/HTTPResponse.hpp
552020: All right! I guess your lobby and me has different views about HTTPResponse and HTTPRequest objects. As far as it is an 'aware' choice I'm fine with it!

Being CGI or not being CGI is quintessential not for being a response but for become the right response for a certain request. The response that reaches the client will have no trace of its being generated through CGI or through a static page. No one will know and the HTTPResponse will also not know it. I can give you two equal in all their quintessential properties http responses (sent by a server): one generated through CGI and one generated through a static file, and you will not be able to tell the difference between them.

src/Connection.cpp
552020: You can do then directly also the easy fix that would make Daniil (and you) happy. So we solve the problem of the parsing core.

src/HTTPResponse.hpp
lmangall: I see your point and I think it is a really good point, it reminds me of environment variables being called meta-variables only inside the CGI (which I don't like)

src/Connection.cpp
lmangall: I don't understand your comment, but I can reverse it if you want, I think I did it at the beginning of the branch and didn't realized it was your part (as our concept of core is a bit flawed)

src/utils.cpp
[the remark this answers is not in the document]
552020: Did you add the cstring header?

```

#include <iostream>
#include <cstring> // For std::strcpy

int main() {
    const char source[] = "Hello, C++98 World!";
    
    char destination[25];

    std::strcpy(destination, source);

    std::cout << "Copied string: " << destination << std::endl;

    return 0;
}
```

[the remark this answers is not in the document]
552020: not string.h the c header but cstring the C++ header.

src/Connection.cpp
552020: Let's leave it as it is, I will check it at a later point. Now I'm aware that it has been changed.

src/utils.cpp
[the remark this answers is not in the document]
lmangall: 🤦 you are right, sorry

src/CGIHandler.cpp
lmangall: Yes but this is more complex, the functions that call the (various) setErrorCode are called all over the place, so I think it is good to have a safeguard, just before (re)writing an error code into the member variable

552020: Do whatever it seems better to you. Maybe write an issue about rewriting status code, which shoudn't happen. Maybe add let the function just return If the statusCode != 0 but I don't know exactly the whole chain.

552020:
I tested the 5 links listed in the description. The first three are working: which is great!
The last two not, but probably they are not valid anymore but I get this:

image

I tested it also in my Ubuntu environment

image

lmangall:
Correct, I moved those to cgi-bin in "development_site"

But did you get a segfault ?

I also have to fix the query string

```
├── var
│   └── www
│       ├── development_site
│       │   └── cgi-bin
│       │       ├── hello.cgi
│       │       ├── hello.cpp
│       │       ├── hello_py.cgi
│       │       ├── hello_var.cgi
│       │       └── thirty_py.cgi
│       ├── perl_site
│       │   ├── about.html
│       │   ├── cgi-bin
│       │   │   ├── calculate.cgi
│       │   │   └── index.cgi
│       │   └── fun_fact
│       │       └── php_not_perl.html
│       ├── php_site
│       │   └── README.md
│       └── python_site
│           ├── cgi-bin
│           │   └── index.py
│           └── index.html
└── webserv
```

src/CGIHandler.cpp
lmangall: I added it then

lmangall:
On my machine I didn't had any segfault, only a leak, but it is solved now

552020:
Is everything done now? Ready to me merged? 

Can we have also localhost:8080/ as valid path?

lmangall:
Done, in accordance with the current state of the discussion at:
https://linear.app/web-serv/issue/WEB-136/decide-on-the-index-of-our-webserv

lmangall:
a non-existing path now serves var/www/errors/404.html

552020:
The homepage is served now! Also the other pages are working and invalid path are served with 404. 🚀

I tried now a simple request with curl: `curl -v http://localhost:8080/` cause the tester I was trying out was not working and the response I got is:
```

*   Trying [::1]:8080...
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.4.0
> Accept: */*
>
* Unsupported HTTP/1 subversion in response
* Closing connection
curl: (1) Unsupported HTTP/1 subversion in response

```

So investigate further and I printed out the response string just before being sent for home

image

And I get this

RESPONSE START
**HTTP/1.1 0 Unknown**
Content-Type: : text/html
[...]

The code should be 200 OK

I don't know if it is relevant for this PR or it should be solved somewhere else...

552020:
Adding the statuscode(200) seems to do the trick

image

lmangall:
Thanks for the testing, with last commit I now have:
```
c4c1c1% curl -v http://localhost:8080/ 
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: : text/html
* Invalid Content-Length: value
* Closing connection 0
curl: (8) Invalid Content-Length: value
```

This is now fine for the 200 OK but I have to investigate for the content length

552020:
Anyway after this I guess we can 'close' this branch, i.e. mark as 'merged' and start to work from here, waiting for Daniil to approve it. The next PR open would be his file upload MR, which is only half of the job and he says that the headers are maybe not parsed correctly cause of their different formatting. 

Maybe we could try to implement and run some testing suite to have a more clear idea where we are. What do you think?

lmangall:
The print might be because I added a printing line, but I don't see it after the curl.

Yes I was also thinking about working on my side of the uploads today, but it is true that we still lack testing

lmangall:
It should be good now, it seems curl prints the content upon succesful response

552020:
Curl is printing the response

lmangall:
by _short reflection pause_ you mean today/ this afternoon ? where we do some testing ?

There is a BBQ (free food!) at 18:00, also we have some food that has to be eaten (lettuce, cheese...)

552020:
> 
> 
> > Curl is printing the response
> 
> 
> 
> This is expected behavior (from curl)
> 
> 
> 
> 
> 
> 

This is what I meant. I was thinking you didn't understand where this print was coming from

552020 (approved):
LGTM!

src/utils.cpp
```diff
@@ -52,35 +36,67 @@ int hexToInt(std::string hex)
...
-bool	isValidErrorCode(std::string errorCode)
+bool isValidErrorCode(std::string errorCode)
 {
-	std::string error_codes[] = {
-        // 4xx Client Error
-        "400", "401", "402", "403", "404", "405", "406", "407",
-        "408", "409", "410", "411", "412", "413", "414", "415",
-        "416", "417", "418", "421", "422", "423", "424", "426",
-        "428", "429", "431", "451",
+	std::string error_codes[] = {// 4xx Client Error
```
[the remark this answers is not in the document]
lmangall: image
This function is Daniils, in my code the coment I have is 5xx Client Error...

lmangall:
@dantol29, just waiting for your review to be pulled

dantol29 (dismissed):
Nice!!! 

File upload from my side is ready, except headers. You can test it from index.html in my PR

552020 (approved):
LGTM

lmangall:
@dantol29 you can accept again, we have tried a few testers and solve some errors, open some requests. We didn't found any good tester though

dantol29 (approved):
OK!

PR #180, opened 2024-04-20
added mux test, added large file, renamed parser test

lmangall:
So this is a minimal testing program for multiplexing capabilities, instruction for the tester are in the file as comments (tested only on mac), the CGI is visibly blocking, the simple GET request are also I believe

I generated a "large" file with :
`dd if=/dev/zero of=largefile bs=20M count=1`
you can create one that is 100M and it is more visible  that it is blocking, if the order of the get request is:

```
      "http://localhost:8080/development_site/largefile_20M",
      "http://localhost:8080", "http://localhost:8080"};
```

The first `Connected. Response code: 200` appears instantly, the second and third takes significant seconds (but 2nd is instantly followed by 3rd)


lmangall:
I added a "chrono" As instance this is the result with the 50mb file :
```
Sending request to http://localhost:8080
Sending request to http://localhost:8080/development_site/largefile
Sending request to http://localhost:8080
Connected. Response code: 200. Time taken: 1.55651 ms
Connected. Response code: 200. Time taken: 8653.32 ms
Connected. Response code: 200. Time taken: 8724.59 ms
```

However, it seems the order in which are served the request is not predictable, or perhaps sometimes it doesn't block... It is strange but on some tests the two classic .html are served first and then the heavy file...

552020:
Now I spent a couple of hours playing around with this and even if the idea of the concurrent requests sent with threads is good, I couldn't figure out a case which could block our server depending on the behaviour of send and recv. The case of the GET request for a large file will block the server, cause we should implement chunked response if the response is bigger than a certain size. I created a new linear issue for it. https://linear.app/web-serv/issue/WEB-180/implement-chunked-body-response-for-large-targets

A POST request to a CGI with a way function will block cause the wait (which needs to be implemented). 

But I couldn't figure out a valid case in which the server would block cause of send and recv as supposed by this test. We handle chunked body and in general we are not blocking on recv. We are blocking on send (first cause we need to implement POLLOUT) and second for big files cause a chunked response is not implemented yet. If you can imagine other cases pleast tell! @lmangall @dantol29

lmangall:
Most likely will be deprecated in favour of one of @552020 's branch with a similar largefile and concurrent threads approach, but without the curl hassle

552020:
How do we want to proceed on this? 

My proposal is: 
- you chechk if this curl thing is of any utility. 
- you check our tests.cpp file and you try to integrate it: the tests there are all sync not async, so we would need a function that take an array of test (like the other ones) and create a thread for each of them. The routine of the thread should send the request (I don't know if we need to create a socket before, but probably we do) and then it should wait for the answer, and check if the answer is the one we expected. In this particular case it will check if the third response being received before the second one. Since the first and the third will hit the root and the second one another resource, we can just check if the first and second response (the content) are the same: in this case the test succeded. 
- Maybe the elapsed time function could be integrated anyway as util in the utils file. 

The result of the test should confirm that the server is 'blocking'. We will send first the whole big file and then send the response for the this request. This is because we send the big file in a single send. And this should not be the case. 

Basically we should send first data for the size of the buffer, return to poll, and then send another size of the buffer. 

If we know beforehand that we are going to server big files (from an hosted website) then we can setup our virtual host accordingly (but these are setting we don't need to implement for the subject), otherwise another strategy to speed up the operation is to increase the size of the client socket (on the server) sending the big file. More hints in this discussion with the friend: https://chat.openai.com/share/a7b5f884-6830-4e43-a363-2043a8c68957

Let me know if you want to do it, or if I should take it over (cause this concerns more core functionalities) or if you want that we do it together.  

@lmangall

552020 (changes requested):

lmangall:
@552020  I can't answer directly to your comment.

I opened [this issue](https://linear.app/web-serv/issue/WEB-216/organize-the-tests)  related but yes I will work some more on this test, thank you for the guidelines

lmangall:
Deprecated in favor of this branch:
https://github.com/dantol29/webserver/pull/260

PR #183, opened 2024-04-21
slombard/web 150 move core parsing functionalities into parser object

552020:
I moved the parsing functionalities from the 'core' (Server/Connection) objects to the Parser. This is only the second step of the refactoring, cause the third steps would be eliminate the duplicates between the old logic of the HTTPRequest object and the old logic of the Server/Connection objects. I already eliminated some duplicates. 

I was thinking I would have managed to reduce `readHeaders`, `readChunkedBody` and `readBody` to a common function `readSocket` but the project failed so that now readSocket reaplace only `readHeaders` and maybe it could be renamed back to readHeaders. 

These are the private properties of Connection before the refactoring

image

These are the private properties left in Connection after

image

Connection as discussed will have besides the actual HTTPResponse also a Parser, a HTTPRequest object, and a `pid` if the from that Connection a new process for the CGI has been started). 

This is still a proposal, cause Daniil should he happy with it (if not we can design it in another way) so I will tag is a Draft (but it would be ready to be merged: it compiles and works).


552020:
Since 130 was approved, now this can be reviewed (and eventually merged). @lmangall @dantol29. It would be nice if you would manage to do a real review and check what I changed. There are many things are not properly nice.

src/Connection.cpp
```diff
@@ -76,131 +52,33 @@ struct pollfd Connection::getPollFd() const
...
-			// In this case we don't want to send an error response, because the client closed the connection
-			return false;
-		}
+		parser.setBuffer(parser.getBuffer() + std::string(buffer, bytesRead));
+		std::cout << "Exiting readSocket" << std::endl;
+		return true;
+	}
+	else if (bytesRead < 0)
+	{
+		perror("recv failed");
```
lmangall, line 73: from the subject:
 _• Checking the value of errno is strictly forbidden after a read or a write operation._

I wonder if  recv is considered "a read operation"...[ I asked the model](https://chat.openai.com/share/6af6870a-6fb9-4e98-8ff9-4498467c30d2) and it answered yes, what do you think ?

lmangall (changes requested):
with Daniil's tester:

- on "body" mode we have mostly errors 499 returned, instead of 400. and the first test, that Daniils marked as 200 valid, is flagged 499 as well

- With the "simple" tests:
This get flagged as "invalid headers" but seems valid:
`POST / HTTP/1.1\r\nHost: www.example.com\r\n\r\n",	   //  200 (OK)
`

- This one passes through the parser, I am not sure if it is valid, probably not (flagged as not valid by Daniil)
` "GET /random HTTP/1.1\r\nHost: www.example.com\r\n\r\n", // 400 (Bad Request)
`

___________________________________


I found a segfault with both
`http://localhost:8080// ` and `http://localhost:8080\\ `
This could be on the router side, or on the parser, perhaps parser as I don't see how this could be a valid address (we rather don't want to have single or double `\ `  `/` on our server, and we also should not accept it in the name of a file @dantol29 
`localhost:8080./../ ` and `localhost:8080./././`  the parser returns our index, which is expected


___________________________________________


Otherwise I went through the code and everything seems clear and clean. I think having an overload operator for debugging prints (that can leave in the repo after being pushed) is better than a print function, but that's a small thing not directly related to our implementation.
Good work!

dantol29:
I didn't even write 499 errors anywhere...

That tester is mostly to check syntax errors, I will fix everything once I come back on Wednesday. @lmangall Can you open an issue related to this problem?

lmangall:
@dantol29 I think this is on @552020 's side now, you should just review it comment on implementation or ask questions for your understanding and report errors,

I think we also don't need issues if we judge these are necessary fixes for the PR to be accepted, we can open an issue if we decide to fix some of theese later (perhaps the //// and \\\in the URI...)

@552020 do you agree ?

552020:
After three hours debugging: I managed to solve something. 
1. There was an error in the new implementation of the Parser. Cause I was calling parseHeaders, before calling parseRequest and the function hasMandatoryHeaders check for the method, which is set by parseRequest. 
2. I modified a little bit the client tester, cause it needs to wait for a response of the Server before shutting down the client, otherwise for the case that the server needs multiple reads to read the full request the connection will be already closed by the client, which will result in an read error. The modified client is in client.cpp and the previous one is in client_original.cpp
3. It seems that the content length in the request, (I debugged only the first request) for the `body` method  has content length of 17 and not 20 as declared. 
OST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Length: 20\r\nContent-Type: "
		"text/plain\r\n\r\nThis\r\nis body\r\n\r\n", 
		
"This" (4), "\r\n" (2), "is body" (7) "\r\n\r\n" (4) = 17

This is also the reason cause I wanted to add a content length checker in the other tester to check the actual content length of the body of the request, before sending a request with a content length not corresponding to the content length of the body. Issue is still open: https://linear.app/web-serv/issue/WEB-122/help-stefano-to-fix-content-length-retrieval-feature-in-daniils-client. 

Wit this two fixes it seems to work.

552020:
I'll explain later the 499 (I added it).

552020:
After three hours debugging: I managed to solve something. 
1. There was an error in the new implementation of the Parser. Cause I was calling parseHeaders, before calling parseRequest and the function hasMandatoryHeaders check for the method, which is set by parseRequest. 
2. I modified a little bit the client tester, cause it needs to wait for a response of the Server before shutting down the client, otherwise for the case that the server needs multiple reads to read the full request the connection will be already closed by the client, which will result in an read error. The modified client is in client.cpp and the previous one is in client_original.cpp
3. It seems that the content length in the request, (I debugged only the first request) for the `body` method  has content length of 17 and not 20 as declared. 
OST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Length: 20\r\nContent-Type: "
		"text/plain\r\n\r\nThis\r\nis body\r\n\r\n", 
		
"This" (4), "\r\n" (2), "is body" (7) "\r\n\r\n" (4) = 17

This is also the reason cause I wanted to add a content length checker in the other tester to check the actual content length of the body of the request, before sending a request with a content length not corresponding to the content length of the body. Issue is still open: https://linear.app/web-serv/issue/WEB-122/help-stefano-to-fix-content-length-retrieval-feature-in-daniils-client. 

Wit this two fixes it seems to work or at least to give other errors.

552020:
For this request: 	`"POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Length: 17\r\nContent-Type: text/plain\r\n\r\nThis\r\nis body\r\n\r\n", // 200 (OK)` now I get this:

image

Is this cause we don't have an www.example.com page @lmangall

lmangall:
> I modified a little bit the client tester, cause it needs to wait for a response of the Server before shutting down the client, otherwise for the case that the server needs multiple reads to read the full request the connection will be already closed by the client, which will result in an read error. The modified client is in client.cpp and the previous one is in client_original.cpp

This is strange, because Daniil's client sends kind of an empty request soo I wouldn't think they are too big, but it's good to know the client doesn't wait on the response. Perhaps our prints are slowing it down (that's what I noticed with the "largefile" tester)

552020:
As extra info how to riproduce this error.
- Pull this branch :-) 
- Start the server
- Recompile tests.cpp and run it with body as SECOND argument!

lmangall:
The message you get with `POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Length: 17\r\nContent-Type: text/plain\r\n\r\nThis\r\nis body\r\n\r\n", // 200 (OK)
`
is because we don't have POST yet

552020:
> This is strange, because Daniil's client sends kind of an empty request soo I wouldn't think they are too big, but it's good to know the client doesn't wait on the response. Perhaps our prints are slowing it down (that's what I noticed with the "largefile" tester)

I guess the problem before was that the declared length was bigger than the real length. So the server was trying to get the whole length and was stuck there. But the client after seding the request shut down without waiting for a response of the server. The server (recv) got that the client shut down the client and then returned an error. But we need to investigate it better, what happens if a client send a request with content length bigger than the real body and then shut down the connection. I'll open a an issue for that.

552020:
> The message you get with POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Length: 17\r\nContent-Type: text/plain\r\n\r\nThis\r\nis body\r\n\r\n", // 200 (OK) 
> is because we don't have POST yet

Ok. So is everything fine on my side? For this request?

And beyond that since we dont' have an example.com website hosted on our server, shouldn't we not respond with an error anyway, cause the resouce will not be there anyway?

552020:
I also added in tests.cpp now in the body array some comment to make it clear why the http request is supposed to be a bad request. Maybe you can check them. And if I'm not wrong the body should not necessarily ends with a CRLF (single or double) and after the double CRLF ending the headers we can have further \r and \n which would be just part of the body. Right? @dantol29

lmangall:
> Ok. So is everything fine on my side? For this request I mean.
>And beyond that since we dont' have an example.com website hosted on our server, shouldn't we not respond with an error anyway, cause the resouce will not be there anyway?

Yes it's fine, the path validation will be done after the method validation

552020:
Timeout is cause we don't send a response back to the client I would say. But we should since they are most 400s already after parsing

552020:
The commit before this comment is solving the problem of the timeout in the first test of the body, add a 501 (method not implemented) if the request is dynamic but not a CGI. Since this is touching the router, check Leo @lmangall if you like it.

dantol29:
1. Parser class looks good, I modified it a bit and added more checks (now it passes all the tests from tests.cpp).

2. My branch (file upload) was merged with main and does not have core parsing functionalities in the Parser class. We have to merge my branch into this one.

3. I like the approach with parseRequest more than calling parseRequestLine and parseRequestHeaders separately. We have to look at this part together tomorrow

552020:
> Parser class looks good, I modified it a bit and added more checks (now it passes all the tests from tests.cpp).

Great!

The proper way to good (to follow best practice standard) would be that you open a branch starting from this one, do your changes and then open a PR from that branch to this branch. If you have never done it, it could be a good exercise, otherwise we can just look at it together tomorrow. 

> My branch (file upload) was merged with main and does not have core parsing functionalities in the Parser class. We have to merge my branch into this one.

Or you can also merge this one (after we change it merging your changes in it) in your file upload branch.

> I like the approach with parseRequest more than calling parseRequestLine and parseRequestHeaders separately. We have to look at this part together tomorrow

At the moment there are repetition. It's like my old logic and your old logic and they need to be 'merged' together. We can look at it together tomorrow and do the whole work of you opening the branch and then opening a MR on this branch together.

lmangall (approved):
LGTM!

552020:
Solve shakespeare! Body read just 1 time! Body+Header bigger than buffer_size no good.

dantol29 (approved):
## PERFECTO!!!

but would be cool to test with big files

lmangall (approved):
LGTM

PR #196, opened 2024-04-25
Daniil/change parser class

dantol29:
## It is ready to be merged into your branch

- I made parseRequestLineAndHeaders() and parseBody() since we pass them separately.
- tests.cpp does not work properly and I could not test the parser
- all changes from file upload are here, so I think we should close file upload PR

### DIFF
--- a/include/webserv.hpp
+++ b/include/webserv.hpp
@@ -21,7 +21,7 @@ std::string toString(const T &value)
 struct File
 {
 	std::map<std::string, std::string> headers;
-	std::vector<std::string> fileContent;
+	std::string fileContent;
 };
 
 bool startsWith(const std::string &fullString, const std::string &starting);
--- a/include/webserv.hpp
+++ b/include/webserv.hpp
@@ -5,6 +5,7 @@
 #include <sstream>
 #include <cstdlib>
 #include <cstring>
+#include <iostream>
 
 template <typename T>
 std::string toString(const T &value)
@@ -13,6 +14,13 @@ std::string toString(const T &value)
 	oss << value;
 	return oss.str();
 }
+
+// for file uploads
+struct File
+{
+	std::map<std::string, std::string> headers;
+	std::vector<std::string> fileContent;
+};
 
 bool startsWith(const std::string &fullString, const std::string &starting);
 std::string readHtml(const std::string &filePath);
--- a/include/webserv.hpp
+++ b/include/webserv.hpp
@@ -6,6 +6,8 @@
 #include <cstdlib>
 #include <cstring>
 #include <iostream>
+#include <map>
+#include <vector>
 
 template <typename T>
 std::string toString(const T &value)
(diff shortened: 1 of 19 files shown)


dantol29:
## After 2 hours Parser correctly saves file upload headers

1. HTML code for file upload added to index.html
_it works!!!!_

2. I had to comment out this part of the Router to allow POST and test it.
```
else if (isDynamicRequest(request))
{
	std::cout << "\033[31mCGI is the only dynamic requests we handle at the moment\033[0m" << std::endl;
	response.setStatusCode(501, "");
}
```
3. Some headers in file upload can have _parametrs_. For example, `Content-Disposition: form-data; name="file"; filename="a"`. Each header and  parameter is saved in the struct File!

4. It was tested wiht browser, but I would add unit tests to see how incorrectly formatted data is handled

552020 (changes requested):
Great job! Thank you! Maybe you can check this:

- Check CRLF in the body
- Check Fileupload
- Check headers with more than 1024 Kb (bigger than buffer size)

But good!

552020:
I'm looking into the ngnix documentation to understand how the config file works and I found also these interesting 'rules' about the size of the headers:

- https://nginx.org/en/docs/http/ngx_http_core_module.html#client_header_buffer_size
- https://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers

lmangall:
@dantol29  Is your PR #130 deprecated in favor of this one ?

552020:
> @dantol29 Is your PR #130 deprecated in favor of this one ?

Is Mallorca so boring? @lmangall 🥱

lmangall:
Yes, and the panoramas and hikes are sub-Teufelsberg standard (Unesco is totally wrong) 🇩🇪, I already miss the cluster

dantol29:
> Check CRLF in the body
Check Fileupload
## Done!

testing with more than 1024 should be a separate issue

dantol29:
@lmangall Yes, PR #130 is deprecated in favor of this one

dantol29:
## I merged your branch and spotted some issues:
1. GET request now does not work (probably because of the POLLOUT). Now it works.
2. It still enters "regular body" after reading headers.

## The good thing is that the second time it goes straight to the readBody, avoiding readHeaders

552020:
> ## I merged your branch and spotted some issues:
> 1. GET request now does not work (probably because of the POLLOUT). Now it works.
Does it work? POLLOUT is not implemented yet ... 
> 2. It still enters "regular body" after reading headers.
This is normal: 'regular body' is printed after the check if it's chunked or regular. It should not print "Entering ReadBody"
> 
> ## The good thing is that the second time it goes straight to the readBody, avoiding readHeaders

552020:
Is file upload now working? Which problem did we have yesterday exactly with Chrome sending sometimes header + body at once sometimes first headers then body?

dantol29:
File upload is working since yesterday. The problem was that headers and body were sent separately sometimes and we cannot reproduce the exact situation.

Also, I wanna finally make our debug messages readable. Now it is a mess and it is hard to find something in the terminal

552020:
Ok. But open another issue and another branch for it otherwise we will never be ready if we keep changing this.

dantol29:
Ok, do you wanna merge this branch?

dantol29:
## 2MB image is uploaded in 1 second!

552020 (approved):
LGTM!

PR #203, opened 2024-04-26
Lmangall/web 168 implement automatic or manual socket fd closure before

lmangall:
I think we still don't have a way to test this, so we can pull it to main in the meantime if it looks fine to you


src/CGIHandler.cpp
```diff
@@ -97,6 +98,17 @@ HTTPResponse CGIHandler::CGIStringToResponse(const std::string &cgiOutput)
 	return response;
 }
 
+void CGIHandler::closeAllSocketFDs()
+{
+	for (std::vector<struct pollfd>::iterator it = _FDsRef->begin(); it != _FDsRef->end(); ++it)
+	{
+		if (it->fd != _pollFd->fd)
```
552020, line 105: This means that you are not closing the one used by the parent process. Why? We discussed this already I guess ...

552020 (changes requested):

552020:
+ conflicts.

src/CGIHandler.cpp
lmangall: Done

dantol29 (approved):
Looks good to me! Would be nice to save child process fd somewhere to make CGI non-blocking

552020 (approved):
LRGTM!

PR #206, opened 2024-04-27
slombard/web 175 fix multiple reads headers and body after poll

552020:
My suggestion on how to solve the problem of the multiple readings. It's not tested. 

I added an extra variable in handleConnection socketHasBeenRead, which is set to false everytime the function is called. ReadSocket will change it to true when the headers are read, and read body will read the body only if the body is not complete and the socket has not been already read. 

I'll also change the name of readSocket into readHeaders (which was the original name). I was thinking I could use just 1 function to read headers and body but it was too complicated to implement it.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -69,9 +69,9 @@ Parser &Connection::getParser()
 	return _parser;
 }
 // Attempts to read HTTP request headers from the client connection into _headersBuffer on the Parser.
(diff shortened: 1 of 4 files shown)


552020:
But I wrote a linear issue for the test. https://linear.app/web-serv/issue/WEB-176/write-a-test-to-check-if-we-do-multiple-reads-without-passing-through

PR #223, opened 2024-04-29
dtolmaco/web 159 refactor config file parsing in c style

dantol29:
## Parser now has ifstream instead of get_next_line + tester for the config file is improved


src/utils.cpp
```diff
@@ -107,20 +107,20 @@ bool isVulnerablePath(const std::string &path)
 	return (false);
 }
 
-int checkFile(const char *path)
-{
-	char buffer[2];
+// int checkFile(const char *path)
+// {
+// 	char buffer[2];
```
552020, line 112: Do we need the commented out functions?

dantol29: I thought Router may need to use it

lmangall (approved):
LGTM !

PR #224, opened 2024-04-29
Slombard/web 153 implement pollout

552020:
This is still not implementing POLLOUT but the preparation for it. 

I divided handleConnection in 4 functions:
- readFromClient (reading and parsing)
- buildResponse (routing and building the response)
- sendToClient (writing)
- closeConnection (closing)

Till now the responses has been sent in multiple places and also closed in multiple places, with this new design we should send and close only in the proper methods. 

The Connection object has three new properties:

- hasFinishedRead
- hadDataToSend
- canBeClosed

The property hasReadSocket which was a bool inside handleConnection has been moved also into the Connection object.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -7,14 +7,9 @@ Connection::Connection(struct pollfd &pollFd, Server &server)
 	_pollFd.fd = pollFd.fd;
 	_pollFd.events = POLLIN;
 	_pollFd.revents = 0;
-	// TODO: should I initialize the _response here?
-	//_response = HTTPResponse();
-	//_request = HTTPRequest();
-	//_parser = Parser();
-	// std::cout << "Connection created" << std::endl;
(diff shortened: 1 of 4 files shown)


dantol29 (approved):
LGTM!

552020:
@dantol29 Thank you. I'm not 100% sure about the logic, so maybe if you can go through (all) the information flow and check if it makes sense it would be good (if you haven't done it yet). :-)

lmangall (approved):
LGTM!
Really clear naming, respecting the convention with "can" and "has"

lmangall:
I'll let you merge and close, just in case you want to do some more before.

552020:
If I change something I guess zou need to re-approve. Let's merge it, and then you can do through the code together when Leo is back, to make it as logical as possible. 

> Logic in the core is very fragile, one wrong if statement, and nothing is working. That is why I am currently trying to write asynchronous python tests to automate this pain

What are async python tests? To automate which pain? The testing pain?

PR #230, opened 2024-04-30
add secretus repo as 'private' submodule

552020:
This PR is related on how to manage 'sensitive' info. I created a private repo to which you've been invited and then added this repo as submodule of this repo, so only you (after you will accept the invitation to that repo) will be able to clone the repo as submodule. In the private repo in the README there is an explanation on how to do this. The private repo has other repos wich are submodules of that repo. This private repo is .gitignored so maybe you will need to use --force to clone it when using the recursive flag

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "secretus"]
+	path = secretus
+	url = git@github.com:552020/webserv_secretus.git


552020:
@lmangall @dantol29 This is 4 days old!

dantol29 (approved):
Looks great! But as I understand we do not have any sensitive information to hide right know?

552020:
> Looks great! But as I understand we do not have any sensitive information to hide right know?

Yes, we don't have 'sensitive' information like passwords and stuff, we just have repos of other people in our repo, which you normally you don't want to have in your repo (above all if it's public).

lmangall (approved):
Great! 
Sorry I missed it !

lmangall:
@552020 I found an easter egg on you .md:

_Certainly! Here's a comprehensive Markdown note for your collaborators_

PR #231, opened 2024-04-30
added default config file

dantol29:
## I looked up how NGINX default configuration looks like and wrote this

> https://gist.github.com/xameeramir/a5cb675fb6a6a64098365e89a239541d

- listen 80
- server_name
- root /var/www/html

```
void ConfigFile::createDefaultFile()
{
	_variables.insert(std::make_pair("listen", "80"));
	_variables.insert(std::make_pair("root", "/var/www/html"));
	_variables.insert(std::make_pair("server_name", "_"));
}
```


552020:
If I read correctly the subject:

_Your program has to take a configuration file as argument, or use a default path._

What they want from us is to have a normal `webserv.config` file, like ngnix has one, whose path we would use if a config file is not passed as argument. The result would be the same though. It's just if we want to follow exactly what the subject says ...

The gist document is interesting, maybe we could save it in the docs.

dantol29:
OK, then it simplifies the problem. We will just have the default path.

I talked with Florian, and he said that the configuration file can have multiple `server{}` parts for different domains. What do you think?

552020:
> OK, then it simplifies the problem. We will just have the default path.
> 
> I talked with Florian, and he said that the configuration file can have multiple `server{}` parts for different domains. What do you think?

I think that this is true. Each server block represents a virtual host (https://en.wikipedia.org/wiki/Virtual_hosting). Each domain name (i.e. server_name should have it's own server block). The different values of server_name are just all possible name of a domain like example.com www.example.com (or the its IP address). Check also this doc from NGNIX: https://www.nginx.com/resources/wiki/start/topics/examples/server_blocks/

Two different websites will have different roots. I don't know if you can reach the same with location, but I guess 1 server block pro website is the way to go.

dantol29:
## Now it is the NGINX way
```
if (argc == 2)
    a.parse(argv[1]);
else
    a.parse(CONFIG_FILE_DEFAULT_PATH);
```

src/main.cpp
```diff
@@ -1,18 +1,27 @@
 #include "Server.hpp"
 #include "ConfigFile.hpp"
 
+# define CONFIG_FILE_DEFAULT_PATH "./config"
```
552020, line 4: The default config file of NGNIX is name `ngnix.conf`, the one of apache `apache2.conf` or `httpd.conf`. Should we not adapt the name of our config file accordingly, like `webserv.conf`?

552020 (changes requested):

552020 (approved):
LGTM

552020 (approved):
LGTM again!

lmangall (approved):
LGTM, having a max URI makes sense but if Nginx  doesn't have it...

PR #232, opened 2024-04-30
added py tests

dantol29:
## Added tests using Python

1. Header size > BUFFER_SIZE (atm return 200)
2. Header size > 8KB (atm return 431)
3. Chunked request (atm return 400)


dantol29:
1 test sends header size more than BUFFER_SIZE, so it should be 200.

2 test sends header size more than 8KB, so it should be 431

3 test sends a chucked request, but our server does not handle them properly atm.

dantol29:
## Added more tests + fixed file upload parsing

1. Test with file upload
2. Test with multiple file uploads
3. After testing I fixed parsing

dantol29:
## Is it OK?

it took more than 5 minutes to upload a 30MB image and at the end it did not work

dantol29:
## Large file is saved successfully if I do not check for CRLF, but in some requests content-length for the file is not provided and I have to know where the file ends and starts the next one. I have to think about it

dantol29:
@552020 I merged main and python tester stopped working...

I rolled back and it seems to work again

dantol29:
@552020 This branch is merged with the main, but the main does not work properly, even just localhost:8080 does not work

552020:
> @552020 This branch is merged with the main, but the main does not work properly, even just localhost:8080 does not work

Then we should open an urgent issue! ;-)

552020:
Are you going to fix it by yourself? Or you need help on this? @dantol29

dantol29:
@552020 I think py tests are correct, the problem is in our webserver

dantol29:
@552020 @lmangall ready to be merged!

dantol29:
@lmangall When I upload multiple files with html I get this error from CGI. Files are saved and parsed correctly, but then it goes to CGI for some reason.

image

lmangall:
Okay, it's pretty strange, only with multiple files ? Is the request differently formed for one or several files ? 
This is a request or router error. When I'll work on the response for an upload I'll check for that in parallel.
Can you open an issue ?

lmangall:
@dantol29  I am going through the PR, this branch is for both py tests AND upload (or large uploads ?), correct ?

dantol29:
@lmangall Yes, for both tests and file upload. While I was testing file upload I found lots of issues and fixed them in the same branch.

src/Server.cpp
```diff
@@ -84,26 +84,32 @@ void Server::startPollEventLoop()
...
-		std::cout << "File created successfully" << std::endl;
+		std::ofstream outfile((it->headers.find("filename"))->second.c_str());
+		if (outfile.is_open())
+		{
+			outfile << it->fileContent;
+			outfile.close();
+			std::cout << "File created successfully" << std::endl;
+		} 
+		else
+			std::cout << "Error opening file" << std::endl;
```
lmangall, line 111: The naming should be changed, in another PR we can set an error, perhaps:
422 Unprocessable Entity: This status code is more specific and indicates that the server understands the request's syntax but cannot process the instructions. This code may be used when the uploaded file is corrupted, lacks a name, or doesn't meet server requirements.

src/Server.cpp
```diff
@@ -84,26 +84,32 @@ void Server::startPollEventLoop()
...
 	{
-		outfile << files.back().fileContent;
-		outfile.close();
-		std::cout << "File created successfully" << std::endl;
+		std::ofstream outfile((it->headers.find("filename"))->second.c_str());
+		if (outfile.is_open())
+		{
+			outfile << it->fileContent;
+			outfile.close();
+			std::cout << "File created successfully" << std::endl;
```
lmangall, line 108: I think we should have a boolean for that, it's an improvement so it can be in a next PR

tests/5mb.jpg
lmangall, line 1: Reminds me of holidays, horrible, so better to be in 42!

tests/general_tests.py
lmangall, line 1: Some comments would have been god with at least a couple comments :)

lmangall (approved):
Please rename these few things and open issue(s) for the improvements

lmangall (changes requested):
Sorry, I meant "Request Changes"

dantol29:
@lmangall I fixed everything you requested + improved py tests a bit!

lmangall (dismissed):
LGTM

552020 (approved):
LGTM!

lmangall (approved):
LGTM

PR #244, opened 2024-05-01
Slombard/web 195 implement pollout but really

552020:
Adding POLLOUT to the flow. You can ask questions. 

Basically, we listen for POLLIN when a socket is created. When we have data to send, we stop listening for POLLIN and we listen for POLLOUT (writability of the socket). When we send the data we switch from listening from POLLOUT back to POLLIN. This will never be used, but for the case that we want to send some response to a chunked body request before the whole request arrived, then it could be useful. It is just the line in writeToClient and it will don't to any arm cause the connection atm is closed anyway. If we implement chunked responses also to send big files then this should not be changed after each writing but we will have a flat writeiscomplete and then we can switch. 

Edit: I added now this property hasFinishedSending, and when we set it to true then we change also the event the socket is listening to.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@
 /secretus/
 /a.out
 **/*.o
+testers/myenv/
+
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -88,7 +88,6 @@
     "ranges": "cpp",
     "span": "cpp",
     "cinttypes": "cpp",
-    "execution": "cpp",
-    "regex": "cpp"
+    "execution": "cpp"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -88,6 +88,7 @@
     "ranges": "cpp",
     "span": "cpp",
     "cinttypes": "cpp",
-    "execution": "cpp"
+    "execution": "cpp",
+    "regex": "cpp"
   }
 }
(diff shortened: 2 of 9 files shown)


dantol29 (approved):
Seems to be right, but is there a way to properly test it? We should make an issue to add tests related to POLLUOT

552020:
> Seems to be right, but is there a way to properly test it? We should make an issue to add tests related to POLLUOT

Yes, definitely. A good test would be a client sending a request and then another one immediately, without closing the connection. However, since we haven't implemented keep-alive, the server might start a new connection for the second request. In this case, a better test might be handling a chunked request where the server starts to respond before the entire request has arrived, even though we haven't implemented this either.

dantol29 (dismissed):
FIXEEEEEEEEEEEEEEEEEEEEEEEEEEEEED!

lmangall (dismissed):
It works !!! LGTM !

dantol29 (approved):
I approve again...

dantol29 (dismissed):
And again...

dantol29 (approved):
Merge-base confilct resolved hopefully

lmangall (approved):
After rebasing

552020:
@dantol29 can you try to accept it again? If it doesn't work I'll open another branch and kind of copy the code

dantol29 (approved):
Thats it!

PR #246, opened 2024-05-01
Slombard/web 204 write tests to check writing times of 100 mb file into a

552020:
MSG_DONTWAIT set on send seems not to replace the functionalites of a non blocking sockets. Some experiments above all for @lmangall


552020:
You don't deserve this.

PR #257, opened 2024-05-04
Fix broken main logic

dantol29:
I started working on the problem, but I need your help!


src/Server.cpp
```diff
@@ -57,11 +57,11 @@ void Server::startPollEventLoop()
...
-					if (_connections[i].getHasFinishedReading() \
-					&& _connections[i].getHasDataToSend())
-						_FDs[i].events = POLLOUT;
-					printFDsVector(_FDs);
-					print_connectionsVector(_connections);
+
+					// it is NOT CORRECT because we do i-- in closeConnection
+					// if (_connections[i].getHasFinishedReading() \
+					// && _connections[i].getHasDataToSend())
+					//_FDs[i].events = POLLOUT;
```
lmangall, line 64: It's good that you just commented out, we might need this in the future.

lmangall (approved):
LGTM, there's a segfault with CGI but that's my part

lmangall:
The problem was: 
In acceptNewConnection, when pushing back the newConnection, `_connections.push_back(newConnection);`  this used the copy constructor, the copy constructor was not copying @552020  new booleans (hasFinishedReading as instance), only initialising them to 0

552020:
> The problem was: In acceptNewConnection, when pushing back the newConnection, `_connections.push_back(newConnection);` this used the copy constructor, the copy constructor was not copying @552020 new booleans (hasFinishedReading as instance), only initialising them to 0

By chances it happened that all the booleans in the new Connections objects should be initialised as false. So this is actualy correct. The copy constructor there is copying a newly created objects with no values, so I don't guess that this was a problem. Even if I noticed a problem with booleans but become true, when they should be false.

dantol29:
@552020 Can you accept so we have working main?

552020:
> @552020 Can you accept so we have working main?

Did you see my questions?

lmangall (approved):
Conflict solved LGTM

dantol29:
> By chances it happened that all the booleans in the new Connections objects should be initialised as false. So this is actualy correct. The copy constructor there is copying a newly created objects with no values, so I don't guess that this was a problem. Even if I noticed a problem with booleans but become true, when they should be false.

@552020 This one? Leo and I tested and found out that copy constructor was the problem. Even know the main does not have the correct copy constructor, but if you add it it is going to work

dantol29:
> > The problem was: In acceptNewConnection, when pushing back the newConnection, `_connections.push_back(newConnection);` this used the copy constructor, the copy constructor was not copying @552020 new booleans (hasFinishedReading as instance), only initialising them to 0
> 
> By chances, it happened that all the booleans in the new Connections objects should be initialized as false. So this is actualy correct. The copy constructor there is copying a newly created objects with no values, so I don't guess that this was a problem. Even if I noticed a problem with booleans but become true, when they should be false.
> 
> Edit (5/5/24) I checked this now in depth again. I built a copy constructor with a macro that control the section with the booleans so that I can copy the booleans or not. And I printed the Connection object just after creation and after it is inserted in the -connections vector with push_back, after insertion I printed the element in the array (also the copy). I tested it with and without explicit assignment to the values of the booleans. As I said yesterday the Connection object at that moment is empty and booleans are initialized to 0 (also false). Probably on your machine it behaves differently, we can check it maybe later why an explicit assignment of the booleans in the copy constructor makes a difference at that point where the Connection object is completely new and the default values would be the same.
> 
> The problem I spotted is not with the copy constructor but with the copy assignment operator when a Connection is removed from the _connections vector (the screenshot of yestarday, but today with more cognition on what is happening). In C++98 in which we don't have the move operator, when the underlying array of a vector is resized basically the objects are copied with the copy assignment operator and since in the copy assignment operator of Connection the booleans are not copied but only the fd from the struct pollfd (also present on the Connection object), when the fd4 is removed (the first client socket connecting the server with the browser which we used to send the response), the fd5 is copied on it, but the values of fd4 for the booleans stays there and only the fd is correctly copied.
> 
> I still don't see through the whole thing and why your fix is working beside the apparently wrong justification. But I hope to reach the solution of the mistery till the end of the day. @lmangall


@552020 I also found it strange that booleans were TRUE, because they should be initialized to FALSE by default. After I changed copy constructor they became FALSE as supposed to. It also does not make sense to me... Maybe because of the unfinished copy constructor something went wrong. I am not sure

552020:
> > > The problem was: In acceptNewConnection, when pushing back the newConnection, `_connections.push_back(newConnection);` this used the copy constructor, the copy constructor was not copying @552020 new booleans (hasFinishedReading as instance), only initialising them to 0
> > 
> > 
> > By chances, it happened that all the booleans in the new Connections objects should be initialized as false. So this is actualy correct. The copy constructor there is copying a newly created objects with no values, so I don't guess that this was a problem. Even if I noticed a problem with booleans but become true, when they should be false.
> > Edit (5/5/24) I checked this now in depth again. I built a copy constructor with a macro that control the section with the booleans so that I can copy the booleans or not. And I printed the Connection object just after creation and after it is inserted in the -connections vector with push_back, after insertion I printed the element in the array (also the copy). I tested it with and without explicit assignment to the values of the booleans. As I said yesterday the Connection object at that moment is empty and booleans are initialized to 0 (also false). Probably on your machine it behaves differently, we can check it maybe later why an explicit assignment of the booleans in the copy constructor makes a difference at that point where the Connection object is completely new and the default values would be the same.
> > The problem I spotted is not with the copy constructor but with the copy assignment operator when a Connection is removed from the _connections vector (the screenshot of yestarday, but today with more cognition on what is happening). In C++98 in which we don't have the move operator, when the underlying array of a vector is resized basically the objects are copied with the copy assignment operator and since in the copy assignment operator of Connection the booleans are not copied but only the fd from the struct pollfd (also present on the Connection object), when the fd4 is removed (the first client socket connecting the server with the browser which we used to send the response), the fd5 is copied on it, but the values of fd4 for the booleans stays there and only the fd is correctly copied.
> > I still don't see through the whole thing and why your fix is working beside the apparently wrong justification. But I hope to reach the solution of the mistery till the end of the day. @lmangall
> 
> @552020 I also found it strange that booleans were TRUE, because they should be initialized to FALSE by default. After I changed copy constructor they became FALSE as supposed to. It also does not make sense to me... Maybe because of the unfinished copy constructor something went wrong. I am not sure

You are checking for this value in the wrong place, i.e. in the new acceptNewConnection function. There the values are false and stay false, no matter how we write the copy constructor cause they are by default false. 

You need to check the values of all Connections object in the closeClientConnection function before and after erase. You will notice that there the values are (still) fucked up. In a normal call to localhost:8080 we will have 3 Connections: 1 Server Connection with fd 3, and two Client Connection with fd 4 and fd 5. We use the Connection 4 to send the response and then we close this connection and the Connection 4 get erased. This means that the connection 5 is copied on the connection 4 with the copy assignment operator and at this point the values get fucked, cause the copy assignment operator was only partially implemented, the fd is correctly copied but the other values are not. 

I built a branch on the top of your branch 214 and I added some prints slombard/web-218-understand-why-the-other-fix-is-actually-working› just to understand why is actually working. I added a lot of prints statements so you can easily check these values.

dantol29:
@552020 What do you mean "removed that check at the end of handleConnection.", I cannot find it.

lmangall:
> There should be something else in the modification you did that fixes main but I don't understand what. Probably the fact that you removed that check at the end of handleConnection.

@552020 I haven't looked yet at your new branch but actually we have both been working on branches where we just added the few lines on the copy constructor (in order not to pull this branch onto another one and have issues with GIT I was told) so it should be those lines that solved the problem...

552020:
> @552020 What do you mean "removed that check at the end of handleConnection.", I cannot find it.

This one:

image

552020:
> > There should be something else in the modification you did that fixes main but I don't understand what. Probably the fact that you removed that check at the end of handleConnection.
> 
> @552020 I haven't looked yet at your new branch but actually we have both been working on branches where we just added the few lines on the copy constructor (in order not to pull this branch onto another one and have issues with GIT I was told) so it should be those lines that solved the problem...

@lmangall It can be. But for sure not in the place where the push_back method is used, which as I told you yesterday make 0 difference to the code, cause these booleans in the new Connection object are false and if not explicitly declared they are also initialized as false in the copy constructor. And sincerely I don't know where the copy constructor is used elsewhere. It can be that the other lines that has been changed made the difference, for the fact that the index `i` can be wrong and POLLOUT could be set as event on a Connection that should not have it. But I need to test this hypothesis.

552020 (approved):
LGTM!

PR #260, opened 2024-05-04
added mux_cgi tester, cleanup testers, added .md tester

lmangall:
I moved everything under the tests folder, added a legacy folder with former version (we can also delete), added my test for non-blocking nature, and a .md with a short description of the tests

I am opening for visibility, the discussion is still open on how to have our tests.

The way I see it now is different programs, when we add one we add a small description on the tests,md, and add it to github actions


552020 (approved):
Awesome!

dantol29 (approved):
OK, I will add my .py tests in this folder.

mux_CGI.cpp does not work because of the main, so I cannot test it

PR #262, opened 2024-05-05
Dtolmaco/web 198 be able to setup multiple server components config file

dantol29:
## Added new class `ServerBlock` that stores data of a server block
image

## `Config` class has a vector with `ServerBlock`
image

I like the way it is organized now!
@552020 @lmangall What do you think?


lmangall (dismissed):
LGTM!

lmangall (approved):
LGTM !

552020 (approved):
Best!

PR #263, opened 2024-05-05
Fix broken main fixing the copy _assignment_ operator (for erase operation in close connection)

552020:
I open this one just for visibility if you are interested in checking on what I'm working on. 

In this question on the other PR which address the same problem is briefly explained what it seems to be the real problem: https://github.com/dantol29/webserver/pull/257#issuecomment-2094203717 but I still don't get exactly why Daniil's fix is actually working.


552020:
This is no more necessary. All improvements have been moved in other PR.

PR #273, opened 2024-05-06
fix logic in the for loop of poll

552020:
Small fix in the logic of the for loop for poll. 

If we add a socket into the _FDs vector we don't want to evaluate that socket before we went to poll again (so we are using the actualSize). But if we remove a socket from the _FDs vector we don't want to use actualSize (which would be bigger) but the size of FDs. 
Basically we loop till to the smaller of the two now.


lmangall (approved):
LGTM!

PR #275, opened 2024-05-06
add formt file to test formatting between me and Leo

552020:
It seems there are some discrepancies in the formatting of the files. Git blame tells that these discrepancies seem to be between @lmangall  and me, I added a file format.cpp with lines that will be formatted differently. I commented the different version out (cause otherwise it would have not be possible to show them). 

We should try to understand from where are they coming from. In the files I added comment telling what gets changed.

### DIFF
--- a/tests/format.cpp
+++ b/tests/format.cpp
@@ -1,0 +1,39 @@
+#include <iostream>
+
+void someFunction()
+{
+	// this line is splitted - Leo has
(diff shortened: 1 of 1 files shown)


lmangall:
Ok, maybe we should look into it in person, because we also have to know relatively to the machine (school computer or our laptops)...

552020:
Kind of solved.

PR #277, opened 2024-05-06
reintegrate old logic to set POLLOUT on the socket

552020:
I reintegrate the old logic. I was thinking it is wrong (following Daniil's idea, who commented it out), but actually it seems to be correct. But maybe I'm wrong again: In the description of the issue I'm explaining why: https://linear.app/web-serv/issue/WEB-227/fix-wrong-logic-to-change-events-to-pollout-which-is-not-wrong. 

I deleted also 
- add comments (cause I don't know which comments we wanted). Daniil please help me to remember
- The snippet is also not so dirty, so I would leave like this. It could be eventually integrate in an extra function.

### DIFF
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -57,13 +57,8 @@ void Server::startPollEventLoop()
 										 _connections[i].getParser(),
 										 _connections[i].getRequest(),
 										 _connections[i].getResponse());
-						// TODO: clean this dirt!
-						// add comments
-
-						// it is NOT CORRECT because we do i-- in closeConnection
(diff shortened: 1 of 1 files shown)


dantol29 (approved):
LGTM! We discussed on Linear why it is correct.

Maybe we can make prints smaller, because 3 lines seems way too many

lmangall (approved):
LGTM!

552020:
> LGTM! We discussed on Linear why it is correct.
> 
> Maybe we can make prints smaller, because 3 lines seems way too many

Opening an issue for that.

PR #280, opened 2024-05-06
Fix mux tester

lmangall:
I fixed my tester which had a couple flaws, you need to test several times (because randomly one request can get before another) 
On my mac when you hit in this order
1 - index
2- Largefile
3- example

Sometimes _Largefile_ blocks the server and _example_ doesn't get served...
Maybe it's because of my machine 

(the largefile is gitignore you need to generate it as explain in the .md in the tests folder)


552020:
Can we have a Makefile for it or a function in the program that generates the big file?

dantol29:
@lmangall needs to fix this tester. I could not properly test anything. (WIP I suppose)

552020:
I hope this test is doing enough low level stuff to justify the use of C++! ;-)

lmangall:
@dantol29  @552020  

We now have :
- a Makefile that creates the largefile (idk if it will be github actions friendly)
- 3 test for the response received
- 1 test for the non blocking (3rd request should be served faster than second (largefile))
- an exit code for the tester to be used with Github actions

On my mac it was non blocking (the 4th test was successful)

lmangall:
I think it will be better to have a largefile (could be of 50mb or 20mb) living in the repo and get rid of the Makefile, what do you think ?

552020:
> Wow! Now it works, but how did you fix the issue in the writeToClient we could not fix yesterday? @lmangall

Yes, they are in the code.

552020:
Can we rename this to something like: implement non blocking trasmission for large file (plust test). I'm not sure we have an issue for that.

dantol29:
## Now it does not pass

image

552020 (changes requested):
Please fix code, so that it passes the test.

552020:
> Can we rename this to something like: implement non blocking trasmission for large file (plust test). I'm not sure we have an issue for that.

@lmangall

lmangall:
@slombard I am not sure where is the problem, we discussed it with @dtolmaco and it is rather the webserv that is failing at non-blocking (but the testing results are random (they shouldn't be, I know))
Mux is a standard short version for multiplexer
I cannot rename the PR, I renamed the branch to `lmangall/implement_non-blocking_transmission_fix_mux-tester`

552020:
Ok. It seems there is still work to do on this, but we can merge it, if you want to have the test in the main branch. Actually it should be merged, when we understand what is not working cause everything is still pretty vague. Is there any reason to merge it, if we are not even sure if the test is working properly? I can take a look at it tomorrow but better when we are done with the multi virtual server part.

lmangall:
>  I can take a look at it tomorrow but better when we are done with the multi virtual server part.

Then it's better to merge this one first, because we are just starting with the multi virtual server 
Then we can come back to this maybe when we do more testing with siege

552020:
> > I can take a look at it tomorrow but better when we are done with the multi virtual server part.
> 
> Then it's better to merge this one first, because we are just starting with the multi virtual server Then we can come back to this maybe when we do more testing with siege

But why merge it if it doesn't work? And anyway you needs to fix history first `git rebase -i main`. If the PM sees this mess of commits he's gonna fire us both! @lmangall

552020:
Conflicts

552020 (changes requested):
Sorry! Tests are failing!

552020:
Sorry! Conflicts!

dantol29:
We have to merge this one ASAP, because it has correct send with BUFFER_SIZE

dantol29:
I noticed if we sent BUFFER_SIZE with send, we load images 1000 times slower...

552020:
> We have to merge this one ASAP, because it has correct send with BUFFER_SIZE

@dantol29  But test are failing, maybe you can cherry pick that commit into another branch: for just that thing.

lmangall:
@552020  @dantol29  I merged main and am solving the conflicts, maybe it'll pass the tests then

dantol29:
@lmangall It is fine. I fixed everything

dantol29:
Everything except 3 case in query string parsr.cpp

dantol29:
@lmangall Your merge does not pass some c++ tests

lmangall:
Mmmmh, the tests are all passing locally, and I also gave no args (therefore webserv_default.conf)

lmangall:
@552020 @dantol29  The tests were not passing because autoindex was on... (therefore I think it was returning a 200 sometimes instead of a 404...) However they did pass localy...
Anyway we can pull to main

dantol29 (approved):
LGTM!

dantol29:
@552020 ASAP

552020:
> @552020 ASAP

An Intern ASAPing the PM? Really?

552020 (approved):
🥗

PR #281, opened 2024-05-06
slombard/web 221 fix copy assignment operator and check the copy constructor

552020:
I added the missing properties in the copy constructor and in the copy assignment operator  of the Connection object, that should fix the bug that you hadn't the time to check - but there was one! I also wrote a copy assignement operator and copy constructor for the HTTPResponse object since it is a property of Connection and the copy constructor and copy assignment operators of HTTPResponse were private. 

Please check that everything is working properly after these changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@ SRCS = src/main.cpp \
 	src/server_utils.cpp \
 	src/Server.cpp \
 	src/StaticContentHandler.cpp \
-	src/Config.cpp
+	src/Config.cpp \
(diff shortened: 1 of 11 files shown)


dantol29 (dismissed):
I checked with python tester and everything works!

We have to integrate it to GitHub Actions!

lmangall (approved):
Great!

552020:
We get again this error: the merge-base changed after approval.

dantol29 (dismissed):
LGTM! REBASEEE

PR #286, opened 2024-05-06
Add type of socket in the connection objects, plus ip, port and implement logic to assign the values

552020:
I tweaked a little bit the Connection object. 

Now Connection objects have a type (UNDEFINED, SERVER, CLIENT), they have an `_ipAddress` property, which is an std::string, and they have an `_ipPort` property. 

The Server Connection objects get this properties from the _serverAddr sockaddr_in struct on the Server. 
The Server Connections get the values from the Server Connection objects that hold the data of the _serverFd which created them. The setup is still for just 1 server socket, but when the SeverBlock vector is ready, so that from each block isit possible to retrieve the ipaddress (as a string) and the port (as unsigned short) then the multi server socket capabilities can be implemented.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -5,6 +5,9 @@ Connection::Connection(struct pollfd &pollFd, Server &server)
 {
 	(void)server;
 	_pollFd.fd = pollFd.fd;
+	_type = UNDEFINED;
+	_serverIp = "";
+	_serverPort = 0;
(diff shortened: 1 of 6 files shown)


dantol29 (dismissed):
Looks good!

dantol29 (approved):
Again, OK!

lmangall (approved):
LR(ebased)TM!

PR #294, opened 2024-05-06
Add awesome prints

552020:
I added in this MR all  the prints I used to spot the error in main. Enjoy!


lmangall (dismissed):
LGTM!

552020:
What is this the merge-base changed after approval?

552020:
Do you think that this is just about the fact that some other PR got merged and then the merge base changed and even it is not conflicting with this one the reviews are considered stale. I don't think we had this before ... I mean is a common practice to approve and merge multiple PR ...

lmangall:
I am not sure, it's quite annoying

src/Server.hpp
```diff
@@ -66,7 +68,7 @@ class Server
 	void listen();
 	/* for startPollEventLoop */
 	void addServerSocketPollFdToVectors();
-	void acceptNewConnection();
+	void acceptNewConnection(Connection &conn);
```
lmangall: What is this change ?

dantol29 (dismissed):
Looks good!

But can you explain acceptNewConnection changes?

src/server_utils.hpp
```diff
@@ -34,5 +34,10 @@ char customToLower(char c);
 void printFDsVector(const std::vector<pollfd> &fds);
 void print_connectionsVector(const std::vector<Connection> &connections);
 void printStrWithNonPrintables(const std::string httpRequest, size_t startPos);
+void printConnections(const std::string &location,
```
lmangall, line 37: We have quite a few prints set by overloading the << operator, wouldn't that be applicable here ?

lmangall:
> Sure! What is not clear?

I added an inline comment on the code about it

552020:
> > Sure! What is not clear?
> 
> I added an inline comment on the code about it

Normally you are supposed to see in the main page of the pull request but I can't see it.

lmangall:
@552020 Sorry, It's because I had not finished the review,

src/Server.hpp
552020: Ah. Before we were taking the _serverFd (as we had only one). This is the preparation for the multi 'server' thing. We are taking the Server Connection Object which has the POLLIN, to pass the information of this server connection to the client connection we are going to create. Maybe it would have been better to call the parameter Connection &ServerConn

552020:
Before 

image


After

image


I thought I changed the message of the only commit in something more meaningful but this didn't work

src/Server.hpp
lmangall: Alright, thanks for the explanation

lmangall (approved):
LGTM!

dantol29 (approved):
I approve after correct merge rebase

PR #299, opened 2024-05-06
reshuffle README

552020:


552020:
Can you check this when you have time ... @lmangall

lmangall:
I think it would be good to add which constraints we had to respect such as _non-blocking without multithreading_ and _90's standards_ And also write about our workflow: linear, github actions (mention the testers)...

552020:
> I think it would be good to add which constraints we had to respect such as _non-blocking without multithreading_ and _90's standards_ And also write about our workflow: linear, github actions (mention the testers)...

Absolutely!

We could have a paragraphs like ... 
## Constraints
- C++98
- CGI (instead of more modern stuff)

Eventually link the subject here

## Workflow / Stack / Tools
 - Git
 - GitHub
 - GitHub Actions
 - Linear
 - Slack
 - ClangFormat
 ## Tests

PR #308, opened 2024-05-07
small test server with two sockets listening to overlapping IPs

552020:
This is just a small test to check the behaviour described in this issue: https://linear.app/web-serv/issue/WEB-245/understing-virtual-servers-with-overlapping-ipport-bindings

The case we have 1 server socket listening to all IPs for a certain port and another socket listeining to a certain ip and a certain port. In this case we have the first listening to all IPs and port 8080 and the second one listening to 127.0.0.1 and port 8080.


552020:
This is now enhanced with a Dockerimage to run it in a Docker also at school where we can't add another localhost IP to test overlapping.cpp. In the README everything is explained.

lmangall:
The Dockerimage is missing

552020:
This can be merged now, I guess. We tested it with Leo yesterday. It is not changing  the code, it just show how Linux behaves if we have two server sockets potentially listening to the same ip:port. The 'specific socket' will be preferred if a request comes to the ip:port to which it is listening to. 

This is of some relevance for the way the Router will pick up the server block and validate if we are listening to that ip:port for a certain server. 

server
{
   listen 
   server_name website1
}

server
{
    listen 127.0.0.1:80
    server_name website2
}

if a request comes to 127.0.0.1:80 for website1 the router should be able to understand that the first block is matching. 
The flow should be like this:
- Router check server_name again host header (it founds the match)
- Router check the _ip and the _port of the matching serverblock, which will be ANY for the ip address and 80 (the default in a normal server) and since it is ANY it doesn't even need to check for the _ip property on the Connection object (to be implemented) and it will check only the port. 
@lmangall

lmangall (dismissed):
LGTM! There were many changes I did in one of my branch but I suppose this is  normal

lmangall:
@552020  conflicts...

552020:
Please approve till before we get more conflicts. @lmangall @dantol29

dantol29 (approved):
LGTM!

lmangall (approved):
LGTM!

PR #309, opened 2024-05-07
Dtolmaco/web 243 improve config file values checks

dantol29:


docs/config_file/config_file.md
552020, line 1: Can we add a link to the NGNIX directives? This is not urgent can be a linear issue (with low priority)
It would be nice also to know which default values would be used if a directive is not defined. 
Should we explain aslo the syntax for locations/

552020 (approved):
LGTM

docs/config_file/config_file.md
dantol29: OK!

Can you look at the changes that I made? What do you think?

docs/config_file/config_file.md
```diff
@@ -23,24 +22,98 @@ server {
...
-- you can get them with std::vector<std::map<std::string, std::string> > getLocations();
-- you can get an error message with std::string getErrorMessage();. If it returns an empty string it means that config file is valid
\ No newline at end of file
+### 1. LISTEN
+- Can be written multiple times per _server block_
+- Specifies the IP address and port on which the server should listen for incoming connections.
+- The format is listen [address]:port.
+- is stored int the `std::vector<std::string>`
+- _OUR PROTECTION_:
+- _DEFAULT VALUE_
```
552020, line 31: What does it mean _OUR PROTECTION_ : _DEFAUL VALUE_ ? What I wanted to know is if the port value is not present we default to 80, if the IP is not present we dafault to localhost and so on. No?

552020: Ok. I understand the protection here. We should maybe check if the port is a valid int, and if the ip address is a valid ip address ...

552020 (commented):
This needs to be approved after the more general one, I guess, right?

docs/config_file/config_file.md
dantol29: I have an issue about it

lmangall (approved):
Nice!

552020:
Can this be reviewed or is still a draft? @dantol29

dantol29:
@552020 It can be reviewed

552020 (changes requested):
1. Commits are not 'Conventional'
2. Can you explain what you did?

dantol29 (commented):
@552020 Commits in the future PRs gonna be written the right way

### Things changed:
1. Added documentation about each directive(we still have  to decide how DEFAULT VALUE should look like if none is provided)
2. Added support for *cgi_path*(`std::string`) and *cgi_ext*(`std::vector<std::string>`)
3. Error page directive is changed to `std::vector<std::pair<int, std::string>>` because it can be written multiple times per server block
4. Listen directive is changed to `std::vector<std::stirng>` because it can be written multiple times per server block(I know that you want it to be a struct). It is probably going to be a struct in the next PR
5. Added duplicate protection for listen directive: `listen 8080 8080` is wrong
6. Added duplicate protection for all directives that can appear only once per server block(almost all of them).
7. Added Directives default constructor(without it had some weird errors)

552020 (approved):
LGTM!

552020:
> @552020 Commits in the future PRs gonna be written the right way
> 
> ### Things changed:
> 1. Added documentation about each directive(we still have  to decide how DEFAULT VALUE should look like if none is provided)
> 2. Added support for _cgi_path_(`std::string`) and _cgi_ext_(`std::vector<std::string>`)
> 3. Error page directive is changed to `std::vector<std::pair<int, std::string>>` because it can be written multiple times per server block
> 4. Listen directive is changed to `std::vector<std::stirng>` because it can be written multiple times per server block(I know that you want it to be a struct). It is probably going to be a struct in the next PR
> 5. Added duplicate protection for listen directive: `listen 8080 8080` is wrong
> 6. Added duplicate protection for all directives that can appear only once per server block(almost all of them).
> 7. Added Directives default constructor(without it had some weird errors)

@dantol29 Thanks! I approved this. 

Regarding 5. Also 
`
listen 80
listen 80
`

on two different lines should be wrong. 

1. Default value, should be atm 8080 (on the deveopment server) and 80 on a real server (like in the droplet). For the IP in listen if not specified it should be all IP addresses.

lmangall (approved):
LGTM!

PR #318, opened 2024-05-08
Lmangall/web 250 implement the use of serverblocks in the router

lmangall:
In this PR:

- [x] the server take a vector of serverblocks
- [x] the router take 1 serverblock
- [x] index for a given location is implemented (_path has been added to request)
- [x] we have a configfile with at least one server block that is taken by the webserv and used throughout the code
- [x] write config files to fit our website directory structure
- [x] isPathValid returns an enum to signal dir listing generation 
- [x] return the error page set in the configfile 
- [x] directory listing 
- [x] the router checks if the server block name matches the host in the request header

In the next PR:
- [ ] in the server we check if the listen in the server block is matching port and ip from connection
=> those settings are not in connection yet 
- [ ] have several virtual server functionnal
- [ ] additionnal checks for directory listing
- [ ] check for several server names
- [ ] allow_methods for given location
- [ ] alias
- [ ] cgi_path
- [ ] cgi_ext


lmangall:
@552020  @dantol29  Now ready to be pulled to main, I updated the PR description as well

src/Router.cpp
```diff
@@ -5,15 +5,35 @@ Router::Router()
...
+Router &Router::operator=(const Router &obj)
+{
+	if (this == &obj)
+		return *this;
+	_serverBlock = obj._serverBlock;
+	_FDsRef = obj._FDsRef;
+	_pollFd = obj._pollFd;
+	return *this;
+}
+
```
552020, line 26: Shoudl we "NULL" the values in the constructors?
Is the copy assignment operator good like this?

src/Router.cpp
```diff
@@ -110,50 +130,66 @@ void Router::splitTarget(const std::string &target)
 	}
 }
 
-bool Router::pathIsValid(HTTPRequest &request, std::string webRoot)
+bool Router::pathIsValid(HTTPRequest &request, std::string &webRoot)
 {
 	std::string host = request.getHost();
+	std::cout << "pathIsValid Host: " << host << std::endl;
+	std::cout << "pathIsValid WebrRoot: " << webRoot << std::endl;
 	size_t pos = host.find(":");
```
552020, line 138: Can we start to use the Debug::log and add a ROUTER enum to print stuff in the Router?

src/Router.cpp
```diff
@@ -110,50 +130,66 @@ void Router::splitTarget(const std::string &target)
...
+	// std::cout << "Host (after : trailing) :" << host << std::endl;
 
-	if (host == "localhost")
-		path = webRoot + path;
-	else
-		path = webRoot + "/" + host + path;
+	// for ease of use during deployment
+	// this if/else allows to reach target with tester or browser
+	// if (host != "localhost")
+	// 	webRoot = host + webRoot;
```
552020, line 148: This was the root of the evil, ritght? And we don't need this anymore. Right?

src/Router.cpp
```diff
@@ -110,50 +130,66 @@ void Router::splitTarget(const std::string &target)
...
+			webRoot += "index.html";
+		}
+		else
 		{
-			path += "/";
+			std::cout << "Index: " << _serverBlock.getIndex()[0] << std::endl;
+			std::string index = _serverBlock.getIndex()[0];
+			webRoot += index;
+			std::cout << "Path: " << webRoot << std::endl;
 		}
```
552020, line 177: You are taking here the first value of getIndex, but we should check for all of them, right? This is a real question cause I don't know exactly how this work, should work.

src/Router.cpp
```diff
@@ -110,50 +130,66 @@ void Router::splitTarget(const std::string &target)
...
-		path += "index.html";
 		// std::cout << "Path: " << path << std::endl;
-		if (stat(path.c_str(), &buffer) != 0)
+		if (stat(webRoot.c_str(), &buffer) != 0)
 		{
+			request.setPath(webRoot);
 			// TODO: decide if we should return a custom error for a directory without an index.html
 			return false;
```
552020, line 183: We don't want to give extra information about the websites we host. If the index is not present and we don't have that directive that allow to listen the content of the directory, then it's a 404 no?

The friend:

The choice between returning a 404 Not Found error and a 403 Forbidden error when a directory is accessed without an index file and directory listing is disabled depends on the intended message and security policy.

- **404 Not Found**: This error implies that the URL points to a resource that does not exist. Using a 404 error might be more user-friendly or less revealing about the server's structure, as it doesn't explicitly indicate that the directory exists. It simply states that the requested resource is unavailable, which could be interpreted as the directory or page not existing at all.

- **403 Forbidden**: This error explicitly tells the user that the server understood the request but refuses to authorize it. If a directory without an index file is accessed and directory listings are disabled, a 403 error is technically correct because the resource (directory listing or index file) is there but access to it is not allowed. This provides a clear indication that the directory exists but the server's configuration prevents its contents from being listed or an index file from being shown.

Choosing between a 404 and a 403 error often depends on the desired security posture and user experience. For example, if you prefer to minimize information about the server’s directory structure and potential points of vulnerability, a 404 might be the better choice. On the other hand, a 403 error might be used when it's okay to reveal the existence of a directory but not its contents, often used in more secure or private areas of a website.

src/Server.cpp
```diff
@@ -89,11 +89,15 @@ void Server::startPollEventLoop()
 void Server::readFromClient(Connection &conn, size_t &i, Parser &parser, HTTPRequest &request, HTTPResponse &response)
 {
 	(void)i;
-	std::cout << "\033[1;36m" << "Entering readFromClient" << "\033[0m" << std::endl;
+	std::cout << "\033[1;36m"
+			  << "Entering readFromClient"
+			  << "\033[0m" << std::endl;
 	// TODO: change to _areHeadersCopmplete
```
552020, line 95: Can we change these with the proper macros and check how to implement this in the Debug::log thing. Maybe linear issue about this?

src/ServerBlock.cpp
552020, line 1: I hope it's the last time I see all these changes cause of the poor synchronisation in our team about autoformatting standards. @lmangall @dantol29

src/StaticContentHandler.cpp
```diff
@@ -56,17 +56,17 @@ void StaticContentHandler::handleRequest(const HTTPRequest &request, HTTPRespons
...
 	// this if/else allows to reach target with tester or browser
-	if (host == "localhost:8080")
-		path = webRoot + requestTarget;
-	else
-		path = webRoot + "/" + host + requestTarget;
+
+	// if (host == "localhost:8080")
+	// 	path = webRoot + requestTarget;
+	// else
+	// 	path = webRoot + "/" + host + requestTarget;
```
552020, line 65: was this comment also in some other object?

var/www/development_site/5mb.jpg
552020, line 1: A 5MB file should not be committed. We can download the file when we need it to test.

var/www/example.com/index.html
552020, line 1: Can you check if www.example.com or example.com is more exact? The var/www/ is a possible directory where the websites are stored. But the actual domain of the website is www.example.com and not example.com, right?

I discussed the issue with the model: https://chat.openai.com/share/1a11553b-cd4d-4b5b-af77-27be8a84f2a6. 
TL;DR: example.com is the original domain, www.example.com is the domain with the website, we coudl have also a ftp.example.com or a smtp.example.com (with not html) or a blog.example.com (with html). Looking forward to ear your stance on it.

var/www/saladbook.sl/html/404_salad.html
552020, line 1: we bought the .xyz domain for the salabbook website. sl is not a supported TLD.

var/www/saladbook.sl/special_effects/.gitkeep
552020, line 1: What is .gitkeep?

552020 (changes requested):
1. History starts to look super professional! 

2. Where did you find the keyword type `add` in `
add(config): add config to match folder structure`. We should stick to ceratin types I guess. The number of types are not regulated. 

https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum
https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

Maybe a decide issue for this?

3. the chore(.gitignores) stuff can be squashed in 1 commit. 
4. Best PR description till now! 🏅
5. Further questions in the code ...

src/Router.cpp
lmangall: Yes that will be in next PR, if the first index is not available, I should check the second, etc...

src/Server.cpp
lmangall: what do you mean by "implement this" you mean Debug printing from which function it was called ?
If so I just checked it and there is no good way for us

src/StaticContentHandler.cpp
lmangall: Yes, before I was doing that across 2 objects, I added a var _path to the request to simplify

var/www/saladbook.sl/special_effects/.gitkeep
lmangall: to keep an empty dir

lmangall: As instance a dire where upload wil go

var/www/example.com/index.html
lmangall: I think the example of  _"www.example.com" and "blog.example.com",_  is a good one because it shows we should have the www. in the name of the directory
I think we should suggest to users that they should give both names to their servers, so it works as a simple redirecting (namecheap offer the option to redirect thos by default)

src/Router.cpp
552020: It is only for development atm ... what do you mean for the parser, 404?

lmangall:
image
@552020  @dantol29  Some of the tests are not up to date anymore with this branch, which is why I added router.cpp in the test dir, as I will continue to implement as per this PR description in the next PR, less tests will pass.
Basically from now on a test should also fit a config file and a dir structure (which is why the way the intra tests are working)

We can pull to main, the PR description got updated

dantol29 (changes requested):
ServerName is a vectpf because we can have multiple Server names. For example: www.example.com and example.com

Can you add your newly created tests to the yaml git actions file.

I will look at what is wrong with Python tests after we approve this PR

dantol29:
Does the test work as expected? I do not see any green check marks

lmangall:
My tests pass with my configfile, I think it is good if the testers keep example.com so I added a webserv_default.conf and now it crashes, overall I started modifying so I will tweak more things... The python test failing because of the connection to the webserver are unrelated

dantol29 (dismissed):
Perhaps the hardest PR so far...
Luckily, Leo managed to pass every test(we have a weird error in python after all tests are passed)
Huge LGTM!

552020 (approved):
LGTM!

dantol29 (approved):
Ok, I found the reason why python tests did not work. A small error in Server.cpp logic...
GitHub Action is always right!

PR #336, opened 2024-05-11
Add multi server sockets capabilities (and fix ServerBlock)

552020:
It's like a huge PR. 

It compiles but I get `Invalid client_max_body_size` if I try to start ./webserv


lmangall:
@552020  I had the same problem, I don't remember what solved it, I think it was the config file.

docs/getaddrinfo.md
```diff
@@ -0,0 +1,139 @@
...
+```cpp
+void ServerSocket::prepareAddress() {
+    struct addrinfo hints, *result;
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = _listen._isIpv6 ? AF_INET6 : AF_INET;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_PASSIVE;
+
+    const char* ip = (_listen._ip.empty() || _listen._ip == "any") ? NULL : _listen._ip.c_str();
+    std::string port = std::to_string(_listen._port);
```
lmangall, line 68: C++ on Linux doesnt likes that.You need to change it for our own toString,

docs/getaddrinfo.md
```diff
@@ -0,0 +1,139 @@
...
+
+- a Listen object, which contains information relative to the listen directive in the server block in the configuration file.
+- a file descriptor that we get from the socket() fuction
+- a serverSocketAddr variable, which is of type sockaddr_storage. This is an agnostic type that could host IPv4 and IPv6 addresses.
+
+...
+
+The `getaddrinfo` function is an essential tool in network programming, especially when setting up server sockets that need to listen on specific ports and IP addresses. This guide provides an overview of how to use `getaddrinfo` for preparing a `ServerSocket` address and the differences between `getaddrinfo` and `inet_ntop`.
+
+Certainly! Let's continue the exposition in the detailed and informative style you've started:
```
lmangall, line 15: Who is this "Certainly!" for ?

docs/getaddrinfo.md
lmangall: I pushed to your branch to solve this and another minor thing

lmangall:
@552020 Tests failing and conflicts

docs/getaddrinfo.md
552020: Yes, I need to review it! It's partially copy/paste. Thanks for reading it!

dantol29:
Does not compile
image

552020:
> Does not compile image

Yes, I fixed this now

dantol29:
@552020  Does not compile again... 
image

dantol29:
I fixed compilation errors and conflicts, but bind() still does not work.
image

552020:
image

It works, I don't know why. The number of the port printed is weird but it seems to connect and communicate. I spent 1 day for a fix which would have required maybe 10 minutes, cause I got list in the Listen struct (now Listen) object. I rewrote Daniil's logic (I don't know why, probably cause I don't understood it). I don't think the new logic is better: it is anyway cumbersome. The only advantage maybe is that we can eventually put some options in the listen string, the way NGNIX has it. 

The bug to fix was related to the fact that we use Dual Stack Socket, i.e. IPv6 sockets that are able also to bind to IPv4 addresses but the point is that these IPv4 addresses can't be passed just like that but needs to be converted to the IPv6 representation of them. And that was it (if it is solved). 

Warning: I'm using some forbidden functions: inet_ntoa, inet_ntop, inet_pton which are utils to convert ip addresses to their binary representation. We can write our inetNtoa, inetNtop and inetPton.

dantol29:
2340 changes... Gonna test it first

dantol29:
@552020 
1. Just `8080` works
2. Does not work...
image
3. `localhost:8080` does not work
4. `[::1]:8080` this works for some reason

552020:
> @552020
> 
> 1. Just `8080` works
> 2. Does not work...
>    image
> 3. `localhost:8080` does not work
> 4. `[::1]:8080` this works for some reason

Works and doesn't work where? If you enter this values in a config files or in the unit test?

dantol29:
@552020 In the config file

dantol29 (approved):
Lots of bugs!!! LGTM!

lmangall (approved):
LGTM!

PR #345, opened 2024-05-12
Waitpid-for-CGI

lmangall:
### Connection Object Properties
- [x] Add `bool hasCGI` to track if a CGI process is associated with the connection.
- [x] Add `int PID` to store the process ID of the forked CGI process.
- [x] Add `bool CGIexpired` to indicate whether the CGI timer has expired.
- [ ] Implement a `CGItimer` using `time_t` or `struct timeval` (or  alarm, getitimer, tmier_create, gettimeofday)to track the start time of the CGI process.
- [x] pass a reference to the `Connection` object through server and router, then add it to the` CGIHandler object`:
```
void Router::routeRequest(Connection &conn, HTTPRequest &request, HTTPResponse &response)
			CGIHandler cgiHandler(&conn);
```

________________________ *under this is a draft* ________________________

### Process Forking and CGI Handling
- [x] Set `hasCGI` to `true` and copy `PID` after forking the process for CGI.
- [ ] Pass the Connection object up to the CGI handler to maintain state across different components.
=> not done, isn' the reference enough ?

### Server Process Monitoring and Polling Adjustments
- [x] Implement `waitpid(-1, &status, WNOHANG)` after `handleConnection` and just before returning to `poll` to manage child processes without blocking.
- [x] Check if `waitpid` returns 0 (no child has exited) or a PID (child has terminated), and handle accordingly.
- [ ] Introduce a variable `int timeout` in the `poll` call, replacing the constant `-1` to adjust based on active CGI processes.

### Server-Wide CGI Tracking
- [x] Add `bool hasCGI` and `int CGIcounter` to the Server object to track active CGI processes.
- [x] Increase `CGIcounter` when a new CGI process is spawned; decrease when a CGI process ends.
- [ ] Adjust `poll` timeout dynamically:
 - [ ] If `hasCGI` is true, set `timeout` to a defined `WITH_CGI` milliseconds.
 - [ ] If `hasCGI` is false, set `timeout` back to `-1` (blocking).

### Connection Lifecycle and CGI Termination
- [ ] When a CGI process PID is returned from `waitpid`, locate the corresponding connection and finalize the CGI response.
- [x] Reset `PID`, `hasCGI`, and other relevant properties after CGI process completion.
- [ ] Optionally close the connection or reset for reuse.
- [ ] Start a CGI timer upon process creation to handle potential timeout.
- [ ] Check for CGI timer expiration (`CGITimerStart + CGI_TIMEOUT`) periodically:
  - [ ] If expired, send a signal to terminate the CGI process and handle client response.

### Safety and Robustness
- [ ] Ensure all properties are appropriately reset even in error conditions or when exceptions occur.
- [ ] Properly handle signals and interruptions during CGI execution and server polling.


lmangall:
@552020 mesage on slack for reference:

> 
> The idea is to store in the Connection object the pid of the process we create when we fork the main process to launch execve.
> The Connection object needs a couple of more properties: bool hasCGI, int PID, a CGItimer (I don’t know the appropriate type), bool CGIexpired. After we fork we set hasCGI to true and we copy the PID. For this the Connection object needs to be passed into the router till to the CGIhandler.
> We need a waitpid(-1, &status, WNOHANG) after the whole handleConnection logic (not were Leo put it), we need to go throug it after the handleConnection, just before returning to poll. waitpid with -1 will wait for all child process and eventually return the pid of the child process which returned. If this is the case.
> If no child has exited, waitpid will return 0 immediately, allowing your program to continue executing without blocking.
> If one of the child processes has terminated, waitpid will return the process ID of the terminated child and provide its exit status through the status pointer (if it’s not NULL).
> When we set hasCGI to true for a connection we want also to change the behaviour of the poll, instead of having `	int ret = poll(_FDs.data(), _FDs.size(), -1);` we will have a variable int timeout for the -1. The Server will have also a property  bool hasCGI, and int CGIcounter . When you set for a Connection hasCGI to true, the CGIcounter get increased (you need to pass a reference to the Server at the same point where you set this, or check it after exiting handleConnection. Anyway the idea is to have this property set to true if some CGI is going on in the Server in general. Right before entering poll, we check if the Server.hasCGI is true, and if it is the case we switch the timeout to the value of the macro WITH_CGI  that will be of some millisecond, this is the frequency we want to go through waitpid. If the Server has no CGI the value go back to -1 and we block on poll
> If the pid of a child process is returned (we loop through the connection vector till we find the connection with that pid and we finish to execute the response of the CGI. We then reset the value of PID and the hasCGI in Connection and all the other values. We are going to close the connection, but if we want to keep the connection alive, is always good to reset the values. We decrease the value of the CGI going on in the server by one.
> The same time that we fork a process and we start the CGI, we let also a CGI timer start. This timer will be a property of the connection object. We take basically a time stamp and we save it. When we pass the waitpid, we loop through all the connection objects, and we check if hasCGI is true, and if this is the case we check if the timer is expired like the saved timestamp + CGI_TIMEOUT that will be between 30 seconds and 1 minute (said the friend). If this is the case, we send a signal to the process and we kill it. And we send a response to the Client.

552020:
Why is this not associated with a linear issue?

src/CGIHandler.cpp
```diff
@@ -8,11 +8,18 @@ CGIHandler::~CGIHandler()
 {
 }
 
+CGIHandler::CGIHandler(Connection *conn)
+{
+	_conn = conn;
+}
+
```
552020, line 15: Is this supposed to be like this?

src/Server.cpp
```diff
@@ -69,6 +167,7 @@ void Server::startPollEventLoop()
 										 _connections[i].getResponse());
 						if (_connections[i].getHasFinishedReading() && _connections[i].getHasDataToSend())
 							_FDs[i].events = POLLOUT;
+						CGIMonitor(_connections[i]);
```
552020, line 170: For me the monitor, i.e. the waitpid needs to be set, when we are done with poll, i.e. the very next line after the if/else conditions checking `ret` the returned value of poll. Probably I said at the end of 'hanldeConnection' in my message, but actually I meant 'just after poll' or right before going back to poll (after poll).

552020 (changes requested):

src/CGIHandler.cpp
lmangall: I added a note somewhere else, it would be better not to have it as a pointer, but I couldn't solve the cross declarations problems...

lmangall:
> Why is this not associated with a linear issue?

Because I wanted it to be fast

src/Server.cpp
552020: I'm reviewing this again, while closing it, trying to understand if it could make sense to waitpid for everyone in the same else branch where we call handleConnection. Without re-checking my solution again I would say that if we are waiting for a CGI to complete we don't want to go necessarly thorough handleConnection is no CGI is returned and to check if it returned we want to go through waitpid, so they are two different things. When CGI is returned we want go back into handleConnection. But maybe you can also thing about it.

src/Server.cpp
```diff
@@ -30,25 +30,123 @@ void Server::startListening()
...
+						}
+						else
+							Debug::log("CGI counter is already at 0", Debug::OCF);
+						// break;
+					}
+				}
+			}
+			else if (WIFSIGNALED(status))
+			{
+				std::cout << "Child exited due to signal: " << WTERMSIG(status) << std::endl;
```
552020, line 120: Maybe you can check if the status handling in my branch is missing something from here. I took just the logic we had in executeCGI. @lmangall

552020:
I'm closing this. The overview in the description is really nice though, maybe you can save it somewhere. And I left some comment in the code about things we could import from this branch and if it makes sense to place waitpid where you placed it @lmangall

PR #360, opened 2024-05-15
This should fix the problem of the multi server thing

552020:
The tests are testing only IPv4 atm, cause I was lazy. It should be pretty easy to add also IPv6. There is a Dockerfile in the /tests/multi_sockets and a Markdown with instructions on how to run it. I wonder if it's working on the computer school, cause this is the purpose of it. 

The PR has more stuff than just that, cause I branched out from the main of my fork, which has more changed in it, but we can create a branch out of it remove what is not need and PR that branch.


.github/workflows/github-actions-demo.yml
```diff
@@ -4,6 +4,7 @@ on:
   push: # Trigger on push events
   pull_request: # Trigger on pull request events
     types: [opened, synchronize] # Trigger on PR opened and PR synchronize events
+  workflow_dispatch: # Allows the workflow to be manually triggered
```
lmangall, line 7: by "workflow" we mean github action tests ?
Is it manual (tests off) at the moment ? And if so why ?

Makefile
```diff
@@ -3,6 +3,13 @@ CXX = c++
 CXXFLAGS = -Wall -Wextra -Werror -std=c++98 -I. -Iinclude -Isrc -g
 DEPFLAGS = -MMD -MP
 
+UNAME_S := $(shell uname -s)
```
lmangall, line 6: That's for your mac, right ?

Can you explain why are you using fsanitize and is it slowing the webserver significantly ?

src/Server.cpp
```diff
@@ -223,10 +215,11 @@ void Server::buildResponse(Connection &conn, size_t &i, HTTPRequest &request, HT
 	std::cout << "Request host: " << request.getSingleHeader("host").second << std::endl;
 	std::cout << "Request target: " << request.getRequestTarget() << std::endl;
 
+	std::string requestTarget = request.getRequestTarget();
 	// if there is "?" in the request target, we need to remove it
-	if (std::find(request.getRequestTarget().begin(), request.getRequestTarget().end(), '?') != request.getRequestTarget().end())
-		request.setRequestTarget(request.getRequestTarget().substr(0, request.getRequestTarget().find("?")));
-	std::cout << "Request target: " << request.getRequestTarget() << std::endl;
+	if (std::find(requestTarget.begin(), requestTarget.end(), '?') != requestTarget.end())
```
lmangall, line 220: Nice

src/Server.cpp
```diff
@@ -309,7 +304,9 @@ void Server::buildResponse(Connection &conn, size_t &i, HTTPRequest &request, HT
 
 void Server::writeToClient(Connection &conn, size_t &i, HTTPResponse &response)
 {
-	std::cout << "\033[1;36m" << "Entering writeToClient" << "\033[0m" << std::endl;
```
lmangall, line 312: Using your color macros will save us from this edge case autoformatter discrepancy

lmangall (commented):
Good job, thank you, I couldn't test the code now, just read it, but hope I will have the time tomorrow

.github/workflows/github-actions-demo.yml
552020: I guess with this line is now on. Or at least this was my intention.

Makefile
552020: I had a segfault (only on Mac), cause of the different way to manage memory among the two systems, this requestTarget thing you commented on. We don't have Valgring on Mac but this fsanitize. We can comment it out, we don't need it actually.

dantol29 (dismissed):
I tested, seems to work! Great! The only thing is that "localhost" is not correctly parsed. Now that we have the correct multi-server thing we can delete ugly `localhost:8080` folder

552020:
> I tested, seems to work! Great! The only thing is that "localhost" is not correctly parsed. Now that we have the correct multi-server thing we can delete ugly `localhost:8080` folder

That 'localhost:8080' folder doesn't let me sleep at night! Happy you see the uglyness! :-)

Cause of the parsing of localhost, I would say that the listen constructor or parser should just translate localhost in an IP Address, it could be a macro or hardcoded. What do you think?

dantol29:
@552020 Yes, I also thought about hardcoding localhost to 127....

I was thinking about multi server thing logic in the Router. Right know we just loop every request through server blocks and try find matching request header == server_name. Is it correct? What if 2 server blocks have the same server_name?

552020:
> @552020 Yes, I also thought about hardcoding localhost to 127....
> 


It's not really 'hardcoding' it, it is more to bring it back to its natural value. Do you remember https://www.youtube.com/watch?v=Pfy4Q-uDV6I&t=402s&ab_channel=ThePrimeTime ... 

> I was thinking about multi server thing logic in the Router. Right know we just loop every request through server blocks and try find matching request header == server_name. Is it correct? What if 2 server blocks have the same server_name?

If two server blocks has the same server_name it is like if you want to serve the same website from two different servers. It can work if you serve part of the website from one server and another one from another server. But I guess this is way too far. But it can be that the two server blocks with the same server_name have different ip:port combinations. In this case you should check from where the request comes from (to which server) and use that server block (also if it is identical to the other one). If they have the same ip:port combination, then you pick the first one. To set two server blocks could maybe make sense if to each server block only a certain amount of resources are allocated and so having two would mean having double of resources. If the first server block (virtual server) is busy the request can be served by the second one. No?

Makefile
552020: We can change the Makefile so that that flag is loaded only if make is started with a special keyword or just comment it out. And we will comment it in, if we need it. Let me know what you think. @lmangall

multi.conf
552020, line 1: At the moment we get 404 from the server if it works. Actually we should get a correct answer. What is incorrect in the server_name and in the root? We could have correct values for server_name and root and then request an invalid page (just to have a 'valid' 404), like requesting through curl www.development_site/non_existing_resource.html @lmangall @dantol29

tests/general_tests.py
552020, line 1: We need some common setting for python and autoformatting. @dantol29

multi.conf
dantol29: try `autoindex on` in the config file. Directories does not have index.html atm

Makefile
lmangall: I see, a keyword would be ideal but commenting in/out is enough.

Did you find fsanitize helpful ? When I was looking for a good Valgrind equivalent, a long time ago, I didn't liked any of the tools. Leaks was a lot of things to add to the code to make it work, fsanitize didn't had precise information...

multi.conf
552020: I'll make an issue then out of this cause I don't know what it means exactly and we don't need to solve it now.

Makefile
552020: Yes, fsanitize was helpful. I managed to fix the leak.

lmangall:
LGTM, should we merge it ?

dantol29:
@552020 @lmangall Yeah, let s merge it!

dantol29 (approved):
Looks good to me!

Makefile
lmangall: Thanks, good to know

lmangall (approved):
LGTM!

PR #362, opened 2024-05-16
Implement non blocking cgi

552020:
If you want to take a look at the code, this is a first draft, it compiles, and most of the logic is there. 

The only part is missing is when the CGI returns successfully (or not). We will not wait in executeCGI anymore  but we will wait in startPollEventLoop: the wait in executeCGI has not been removed yet. After the CGI returned successfully in startPollEventLoop, where we wait for all processes (-1) without blocking (WNOHANG), we go back to poll, the event we listen for for that connection will be POLLOUT. And then we need a flag like CGIhasFinished. We already collect the exitStatus of the CGI which is set in the Connection object. 

The code to read the buffer written by the CGI is in handleConnection > buildResponse > routeRequest > handleRequest > executeCGI.  We should re-enter handleConnection and buildResponse, but skip the whole logic we already executed the first time we entered it to execute the CGI. Alternatively we could also pass the fd of the pipe when we execute 	_connection.addCGI(pid); but this is dirtier. 

I don't know if you @lmangall has capacity to think about it. I guess with a flag like _CGIhasTerminated (which means it has been executed or kill) we could skip all the unnecessary part of the code, and come to the point where we check the status and read from the buffer. Probably the first step should be just split executeCGI in two function: executeCGI and processExecutedCGI (or something like this), where we read from the buffer or depending on the status code we send back some error.


dantol29:
@552020 It does not compile according to GitHub actions Ubuntu

lmangall:
@552020 Just to know where we stand : Is this code an enhanced version of of this branch ? 
https://github.com/dantol29/webserver/pull/345

include/webserv.hpp
```diff
@@ -12,8 +12,13 @@
 #include <algorithm>
 #include <sys/socket.h>
 #include <netdb.h>
+#include <cstdio>
 #include "Debug.hpp"
 
+#define SEND_BUFFER_SIZE 1024 * 100 // 100 KB
+const int BUFFER_SIZE = 1025;
```
dantol29, line 19: Why do we need a global variable?

src/CGIHandler.cpp
```diff
@@ -1,53 +1,47 @@
...
 	env.HTTPRequestToMetaVars(request, env);
-	// std::cout << env;
-	std::string cgiOutput = executeCGI(env);
-	// if cgioutput == "HTTP/1.1 500 Internal Server Error\r\nContent-Length: 0\r\n\r\n";
-	// then we should return 500 error
-	if (cgiOutput == "HTTP/1.1 500 Internal Server Error\r\nContent-Length: 0\r\n\r\n")
+	if (!executeCGI(env, response))
 	{
 		response.setStatusCode(500, "");
 		response.setBody("500 Internal Server Error");
```
dantol29, line 43: Do we go through Router after this? User may provide custom 500 error page

src/CGIHandler.cpp
```diff
@@ -176,49 +133,35 @@ std::string CGIHandler::executeCGI(const MetaVariables &env)
 		if (access(argvPointers[0], X_OK) == -1)
 		{
 			perror("access");
+			return false;
 			_exit(EXIT_FAILURE);
+			// TODO: @leo I don't think we should exit here. We don't want to kill the whole server cause of a CGI
```
dantol29, line 138: Why don't we set an appropriate error message and error code here? Something like `access denied`

src/Connection.hpp
```diff
@@ -38,6 +38,11 @@ class Connection
 	size_t _responseSize;
 	size_t _responseSizeSent;
 	std::string _responseString;
+	bool _hasCGI;
+	bool _CGIHasExited;
+	pid_t _CGIPid;
+	int _CGIExitStatus;
+	time_t _CGIStartTime;
```
dantol29, line 45: Maybe it would be better to make a CGI struct, but it is a small design thing.

src/HTTPResponse.hpp
```diff
@@ -25,12 +25,28 @@ class HTTPResponse
...
+	void setIsCGI(bool value);
+
+	// Easy version
+	// int *getCGIpipeFD();
+	// void setCGIpipeFD(int *CGIpipeFD);
+	// Fancy version
+	// int (*getCGIpipeFD())[2];
+	// void setCGIpipeFD(int (*CGIpipeFD)[2]);
+	// Fancy version with reference
+	int (&getCGIpipeFD())[2];
```
dantol29, line 42: How it this monster called?

src/Server.cpp
```diff
@@ -41,12 +60,17 @@ void Server::startPollEventLoop()
 {
 	addServerSocketsPollFdToVectors();
 	int pollCounter = 0;
+	int timeout = -1;
 	while (1)
 	{
+		if (_hasCGI)
+			timeout = 500;
```
dantol29, line 67: Does it mean that CGI can be executed max 500ms?

src/Server.cpp
```diff
@@ -213,103 +282,171 @@ void Server::readFromClient(Connection &conn, size_t &i, Parser &parser, HTTPReq
 void Server::buildResponse(Connection &conn, size_t &i, HTTPRequest &request, HTTPResponse &response)
```
dantol29, line 282: We definitely have to split up this function

src/main.cpp
```diff
@@ -14,11 +18,14 @@ int main(int argc, char **argv)
...
 	if (!config.getErrorMessage().empty())
 		return 1;
-	
-	//std::cout << config << std::endl; // should be in the DEBUG?
-	Debug::enable(true);
-	Debug::setLevel(Debug::NORMAL);
-	Server webserv(config);
+
+	std::cout << config << std::endl; // should be in the DEBUG?
+	EventManager eventManager;
```
dantol29, line 23: Can you explain this `event manager`

dantol29 (changes requested):

include/webserv.hpp
552020: We had it already, I just moved it from wherever it was, because of problem of dependencies. But I see what you mean now. I'll tranform it in a macro.

552020: Ok. This should be solved now.

src/CGIHandler.cpp
552020: After this we go through the router (but we skip it kind of). I wrote a line that if response.getIsCGI is skipping the part of the router that we already went thorough before accessing executeCGI. We already went through the Router before this and all the settings should be saved in Connection. I hope they are saved there.

src/CGIHandler.cpp
552020: We can. I'm not the master of error codes. Tell me what is the correct error code. Are we checking here the interpreter or the requested script? @lmangall If you tell me which error should be set I can just change it.

src/Connection.hpp
552020: I took a note: https://linear.app/web-serv/issue/WEB-274/create-a-cgi-struct-inside-connection

src/HTTPResponse.hpp
552020: Array-Referencing Function Beast!

src/Server.cpp
552020: Nope. This means we wait 500 ms on poll and then we go back to waitpid, we pass through poll. We wait 10 long seconds for the CGI to complete. This is another variable or macro.

src/CGIHandler.cpp
dantol29: So this is not correct

src/HTTPResponse.hpp
dantol29: Damn...

src/CGIHandler.cpp
552020: Why not? Normally in the previous code we were not going through the Router again after this. Now we are.

552020:
This needs to be accepted yet. @lmangall @dantol29

dantol29 (approved):
LGTM! Did you test after I left?

552020:
> LGTM! Did you test after I left?

We were testing in the main therefore it was failing. But I didn't test the branch, soon to be main.

lmangall:
Nice, great job !

lmangall:
Did you test with the "parallel CGI" test ?

552020:
> Did you test with the "parallel CGI" test ?

I tried but we have some issues: https://linear.app/web-serv/issue/WEB-330/cgi-bug-in-main-%F0%9F%9A%A8

PR #366, opened 2024-05-17
fix(HTTPRequest): small overflow on mac in setRequestTarget if the target is an empty string

552020:
I also added in the defaut config file a server block for our php_site.


dantol29 (dismissed):
OK!

552020:
@lmangall

dantol29 (approved):
small one! lgtm!

lmangall (commented):
LGTM

PR #368, opened 2024-05-17
Fully fledged PHP website - WIP

552020:
I'm sharing this cause I didn't manage to run it with an index.php file and some php going on that page. Our previous PHP site was just html with no php. 

In this branch I already added @dantol29 last branch on build response. Can one of you router expert  tell me why this is not working. 

In the configuration file I wrote
index index.php
cgi_path (with the path to the executable of the php interpreter)
cgi_ext .php

If you want to test you need to change the path to the executable of the php interpreter


dantol29:
I fixed the Router a bit. It should work. However, index.php is downloaded as file... @lmangall Do you know what may be the issue

lmangall:
@dantol29 regarding the accidental downloading of the .php file
- If this wasn't changed the path to interpreter was a macro, the config file directive was not used.
- you would have to test in the Dockerfile Stefano prepared because we might not have a php interpreter on our machines (or brew install it).

I'll test it as well

552020:
> @552020 I am working on mostly POST, DELETE for the CGI with the script (database like with a json) we already have in main. I pass the data from the body through a second pipe for the stdin of the CGI, now exactly I am working on reading and saving the body of the request for a POST & CGi request
> 
> Are you doing something similar here ?

There will be a major change in main. We are kind of overwriting it cause merge main into the branch with the working CGI was causing to many problems. 

This branch is only about writing a PHP website and should not change the logic of the server itself.

PR #371, opened 2024-05-18
Slombard/web 312 add ipv6 addressees to the multi server test

552020:
This has tests now also for IPv6. 
IPv6 needs to be enabled in Docker though. 
In the Markdown the are instructions on how to do it. 

We have 4 different scripts. 3 out of 4 are working, the last not, cause the config parser is not happy with the one with the mixed addresses with the same port. Anyway the server is definetevly working with multiple ip:ports combination (with different server sockets) for Ipv4 and IPv6. 

IPv6 was not even part of the subject btw.

### DIFF
--- a/multiIPv4+IPv6.conf
+++ b/multiIPv4+IPv6.conf
@@ -1,0 +1,20 @@
+server {
+	listen 0.0.0.0:8081;
+	listen [::]:8081;
+	server_name www.development_site;
+	root /var/www/html;
+}
+
+server {
+	listen 127.0.0.1:8080;
+	listen [::1]:8080;
+	server_name www.saladbook;
+	root /var/www/html;
+}
+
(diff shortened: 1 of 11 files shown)


dantol29 (commented):
I suppose we have to fix `Duplicate listen directive` error message?

552020:
> I suppose we have to fix `Duplicate listen directive` error message?

This is tricky. The subject or the eval sheet say it should not work if you listen to the same port. But different IP and same port make totally sense, like if you have the Ethernet cables connected with your computer, and you want to listen for a certain websites only to two of them. You'll have 2 listen in the same server_block with differen IPs and same port. No?

dantol29 (approved):
:ok_person: I tested some of those! Seems to work!

PR #372, opened 2024-05-19
Add test to check parallels cgis

552020:
The working script is tests/parallel_cgi/duration_ts.sh. 

It sends 3 curls requests to www.development_site/cgi-bin/duration_ts.sh. 
This is a python script the returns an html site with the starting time and the ending time of the script. 
The bash script wait for the answers, collect them, and sort them so that if see if all three scripts start before the previous one end. 
On the actual main it works, but the cgi execute the second script after finishing the first and so on. 
I'll try now this on the branch with the non blocking cgi, but this tests works independently


dantol29 (approved):
LGTM!

PR #373, opened 2024-05-19
Prevent firefox to send background requests to services

552020:
Some configuration file and docs to avoid firefox to send all these background requests.

I found also an interesting article about it: https://www.secureideas.com/blog/2018/10/silencing-firefoxs-chattiness-for-web-app-testing.html

Firefox should then not be used for normal browsing with this config.


552020:
Yohoo! Free merge! @dantol29 @lmangall

PR #380, opened 2024-05-20
fix(segfault): solve buffer overflow due to out-of-bounds access.

lmangall:
Main segfaults, doesn't it on your machine as well ?

Each loop from the overloading operator ant the end of server.cpp now has its own counter variable to avoid variable shadowing.


552020 (approved):

552020:
I opened it for you now.

lmangall:
Merged and deleted the new one

PR #383, opened 2024-05-21
Lmangall/web 270 implement post and delete for cgi

lmangall:
- [ ] Test script out of webserver, with data in query string
- [ ] Modify script to use methods instead of ?add in the query string
- [ ] Add a pipe so the body request is sent to the CGI through stdin (per subject and convention)
- [ ] Add an "early routing" for a POST & CGI request (will need to readBody in parser then go back to server)
- [ ] 
- [ ] 
- [ ] Cleanup Server.cpp to take off the function copied from router or other

Current problem: body is read by the parser but I can't retrieve it in the request

### DIFF
--- a/src/CGIHandler.cpp
+++ b/src/CGIHandler.cpp
@@ -60,6 +60,8 @@ std::vector<std::string> CGIHandler::createArgvForExecve(const MetaVariables &en
 	std::cout << "createArgvForExecve: pathTranslated: " << pathTranslated << std::endl;
 	std::string scriptPath = pathTranslated;
 	std::cout << "createArgvForExecve: scriptPath: " << scriptPath << std::endl;
+	std::string queryString = env.getVar("QUERY_STRING");
+	std::cout << "createArgvForExecve: queryString: " << queryString << std::endl;
 
 	if (env.getVar("X_INTERPRETER_PATH") != "")
 	{
@@ -71,6 +73,7 @@ std::vector<std::string> CGIHandler::createArgvForExecve(const MetaVariables &en
 	{
 		argv.push_back(scriptPath);
 	}
+
 	return argv;
 }
 
--- a/src/CGIHandler.cpp
+++ b/src/CGIHandler.cpp
@@ -37,7 +37,16 @@ void CGIHandler::handleRequest(HTTPRequest &request, HTTPResponse &response)
 {
 	MetaVariables env;
 	env.HTTPRequestToMetaVars(request, env);
-	if (!executeCGI(env, response))
+
+	std::cout << std::endl;
+	std::cout << std::endl;
+	std::cout << std::endl;
+	std::cout << request << std::endl;
+	std::cout << request.getBody() << std::endl;
+	std::cout << std::endl;
+	std::cout << std::endl;
+
+	if (!executeCGI(env, request.getBody(), response))
 	{
 		response.setStatusCode(500, "");
 		response.setBody("500 Internal Server Error");
@@ -101,16 +110,27 @@ void handleTimeout(int sig)
 	std::cout << "CGIHandler: Timeout" << std::endl;
 }
 
-bool CGIHandler::executeCGI(const MetaVariables &env, HTTPResponse &response)
+bool CGIHandler::executeCGI(const MetaVariables &env, std::string body, HTTPResponse &response)
 {
 	std::string cgiOutput;
 	std::vector<std::string> argv = createArgvForExecve(env);
 	std::vector<std::string> envp = env.getForExecve();
 
+	Debug::log("CGIHandler: executeCGI: body: " + body, Debug::NORMAL);
+
 	int pipeFD[2];
+	int bodyPipeFD[2];
 	if (pipe(pipeFD) == -1)
 	{
 		perror("pipe failed");
+		return false;
+	}
+
+	if (pipe(bodyPipeFD) == -1)
+	{
+		perror("body pipe failed");
+		close(pipeFD[0]);
+		close(pipeFD[1]);
 		return false;
 	}
 
@@ -120,6 +140,8 @@ bool CGIHandler::executeCGI(const MetaVariables &env, HTTPResponse &response)
 		perror("fork failed");
 		close(pipeFD[0]);
 		close(pipeFD[1]);
+		close(bodyPipeFD[0]);
+		close(bodyPipeFD[1]);
 		return false;
 	}
 	else if (pid == 0)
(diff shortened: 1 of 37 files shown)


src/CGIHandler.cpp
```diff
@@ -103,26 +110,39 @@ void handleTimeout(int sig)
 	std::cout << "CGIHandler: Timeout" << std::endl;
 }
 
-bool CGIHandler::executeCGI(const MetaVariables &env, HTTPResponse &response)
+bool CGIHandler::executeCGI(const MetaVariables &env, std::string body, HTTPResponse &response)
```
552020, line 113: const std::string body maybe? Or are we gonna change it?

src/CGIHandler.cpp
```diff
@@ -103,26 +110,39 @@ void handleTimeout(int sig)
...
 {
 	std::cout << RED << "Entering CGIHandler::executeCGI" << RESET << std::endl;
 	std::string cgiOutput;
 	std::vector<std::string> argv = createArgvForExecve(env);
 	std::vector<std::string> envp = env.getForExecve();
 
+	Debug::log("CGIHandler: executeCGI: body: " + body, Debug::NORMAL);
+
 	int pipeFD[2];
+	int bodyPipeFD[2];
```
552020, line 123: I don't know a lot about the CGI, but I was thinking that the body is getting "into" the CGI process through the STDIN. Why do we need two pipes? The main process write the body in the fd 1 of the pipe, which becomes the STDIN in the child, the CGI read (automatically) form STDIN. And output to STDOUT which is set as the read end of the same pipe from which the body went in. No?

552020: Discussed this with the friend. You are right, we need two pipes. Sockets are bi-directional, Pipes are uni-directional.

src/MetaVariables.cpp
```diff
@@ -131,29 +131,15 @@ void MetaVariables::subtractQueryFromPathInfo(std::string &pathInfo, const std::
 
 void MetaVariables::HTTPRequestToMetaVars(const HTTPRequest &request, MetaVariables &env)
 {
-	// TODO: several metavars have to be set from configfile
-
-	// This line will have the code put the interpreter path as argv[0] to execve
-	//  env.setVar("X_INTERPRETER_PATH", "/home/lmangall/.brew/bin/python3"); // school computer...
 	env.setVar("X_INTERPRETER_PATH", X_INTERPRETER_PATH);
-
```
552020, line 139: Since i'm reading this. Even if it's part of another Issue, you can check if when we set cgi_path this meta variable is set correctly and if execve is set with the right stuff.

src/MetaVariables.cpp
```diff
@@ -188,14 +167,15 @@ void MetaVariables::HTTPRequestToMetaVars(const HTTPRequest &request, MetaVariab
 	{
 		env.setVar("CONTENT_TYPE", "");
 	}
-	std::pair<std::string, std::string> contentLengthHeader = request.getSingleHeader("Content-Length");
+	std::pair<std::string, std::string> contentLengthHeader = request.getSingleHeader("content-length");
 	if (!contentLengthHeader.first.empty())
 	{
 		env.setVar("CONTENT_LENGTH", contentLengthHeader.second);
+		Debug::log("Content-Length header found.", Debug::NORMAL);
```
552020, line 174: We have a CGI value for Debug now.

src/Parser.cpp
```diff
@@ -223,7 +223,7 @@ void Parser::parseFileUpload(const std::string &request, HTTPRequest &req, HTTPR
 
 // ----------------UTILS----------------------------
 
-bool Parser::hasMandatoryHeaders(HTTPRequest &req, HTTPResponse& res)
+bool Parser::hasMandatoryHeaders(HTTPRequest &req, HTTPResponse &res)
```
552020, line 226: Daniil is doing this on purpose. :-(

src/Server.cpp
```diff
@@ -221,21 +221,21 @@ void Server::readFromClient(Connection &conn, size_t &i, Parser &parser, HTTPReq
...
-	if (request.getMethod() == "GET" || request.getMethod() == "DELETE" || request.getMethod() == "SALAD")
+	if (request.getMethod() == "GET" || request.getMethod() == "SALAD")
 		Debug::log("GET request, no body to read", Debug::NORMAL);
 	else
-		handlePostRequest(conn, parser, request, response);
+		handlePostAndDelete(conn, parser, request, response);
 }
 
-void Server::handlePostRequest(Connection &conn, Parser &parser, HTTPRequest &request, HTTPResponse &response)
+void Server::handlePostAndDelete(Connection &conn, Parser &parser, HTTPRequest &request, HTTPResponse &response)
```
552020, line 230: You can streamline this taking only connection as argument, and retrieving a reference to parser in the function itself:

Parser &parser  = conn.getParser();

I had also this kind of logic in handleConnection and we stramlined this other way, but it's more an aesthetic choice.

552020 (changes requested):
🥗

dantol29:
Less avoid std:cout, we already have working Debug class

src/Server.cpp
```diff
@@ -221,21 +221,21 @@ void Server::readFromClient(Connection &conn, size_t &i, Parser &parser, HTTPReq
...
 {
 	if (parser.getIsChunked() && !conn.getHasReadSocket())
 	{
 		Debug::log("Chunked body", Debug::NORMAL);
-		if (!conn.readChunkedBody(parser))
+		if (!conn.readChunkedBody(parser) && (!conn.readBody(parser, request, response)))
```
dantol29, line 235: Why if the request is chucked, we read in a normal way?

dantol29:
@lmangall I gonna polish your branch a bit because we won t have much time when you arrive

lmangall:
@dantol29 I saw I am writting to you on slack

src/Server.cpp
lmangall: @dantol29  Sorry this is an error, I went through this function (add some point had a copy of it) to identify a problem and I didn't knew it well so I modified something I shouldn't have

dantol29:
@lmangall This is response from DELETE CGI:
```
-----------------CGIStringToResponse-------------------
Entering writeToClient
HTTPResponse operator<< called
Status Code: 200
Headers:
Content-type: text/html

Body:


Status: 400 Bad Request

Content-Type: text/html



<html><body>{"error": "Invalid request. Make sure you provide name and salad parameters for adding, or name for deleting."}<script>console.log("Database filename is /home/dtolmaco/Desktop/webserver/var/www.saladbook.xyz/database/database.json");</script><script>console.log("Attempting to load database from /home/dtolmaco/Desktop/webserver/var/www.saladbook.xyz/database/database.json");</script><script>console.log("Loaded data: {'Stefano': 'Carrot', 'Leo': 'nicoise', 'Daniil': 'with_beets', 'Jane': 'Caesar'}");</script><script>console.log("Received DELETE data: ");</script><script>console.log("form_data={}");</script><script>console.log("method=DELETE, name=None, salad=None");</script><script>console.log("Invalid request. Make sure you provide name and salad parameters for adding, or name for deleting.");</script></body></html>
```

PR #390, opened 2024-05-23
Slombard/web 171 implement https

552020:
I'll open this only for visibility. It has a batter version of the Debug::log function that takes also color as (optional) option. @dantol29


dantol29:
Does not compile according to git actions

552020:
> Does not compile according to git actions

Now it's solved.

552020:
I mean. It compiles. Https is not implemented yet. The problem was that openssl needs to be installed as dependencies into the container of Github Actions.

552020:
Another couple of hours for another small step forward. More tomorrow. Not so far from the end.

